Issue #3, opened 2018-10-23
Enable Buildkite CI


robcohen:
Enable buildkite CI and enable the pipeline.

robcohen:
Removed nixpkgs related files, CI appears to be working correctly now.

PR #4, opened 2018-10-23
#3 - Buildkite nix pipeline added

robcohen:
Tests appear to be passing, github settings need to be changed to check for completed compilation


robcohen:
Apparently, buildkite check is already activated.

release.nix
```diff
@@ -0,0 +1,25 @@
+let
+
+  config = {
+    packageOverrides = pkgs: rec {
+      haskellPackages = pkgs.haskell.packages.ghc822.override {
```
ksaric: Wrong GHC version.

release.nix
```diff
@@ -0,0 +1,25 @@
...
+        overrides = haskellPackagesNew: haskellPackagesOld: rec {
+          universum           = pkgs.haskell.lib.dontCheck (haskellPackagesNew.callHackage "universum" "1.1.0" {});
+          # LIB
+          cardano-shell  = haskellPackagesNew.callPackage ./cardano-shell.nix { }; 
+          cardano-chain-env = pkgs.haskell.lib.dontCheck (haskellPackagesNew.lib.buildStackProject {
```
ksaric: We don't need this and I have no clue how this is supposed to work like this.

release.nix
```diff
@@ -0,0 +1,25 @@
+let
+
+  config = {
+    packageOverrides = pkgs: rec {
+      haskellPackages = pkgs.haskell.packages.ghc822.override {
+        overrides = haskellPackagesNew: haskellPackagesOld: rec {
+          universum           = pkgs.haskell.lib.dontCheck (haskellPackagesNew.callHackage "universum" "1.1.0" {});
+          # LIB
+          cardano-shell  = haskellPackagesNew.callPackage ./cardano-shell.nix { };
```
ksaric: `=` isn't aligned-

scripts/nix/fetchNixpkgs.nix
```diff
@@ -0,0 +1,43 @@
+{ owner
```
ksaric: We don't need this file.

scripts/nix/nixpkgs-src.json
```diff
@@ -2,5 +2,6 @@
   "owner": "NixOS",
   "repo": "nixpkgs",
   "rev": "4dd9cd3f69e51dcb69d29a88335f06a544b0e508",
-  "sha256": "0spbpy4c4mvylvdaiic7daw8lfhd4w2mbqx0q2ywmxpqj5hxnhqr"
+  "sha256": "0spbpy4c4mvylvdaiic7daw8lfhd4w2mbqx0q2ywmxpqj5hxnhqr",
```
ksaric: We don't need this.

stack.yaml
```diff
@@ -65,5 +65,5 @@ packages:
 # Allow a newer minor version of GHC than the snapshot specifies
 # compiler-check: newer-minor
 nix:
-  shell-file: scripts/nix/stack-shell.nix
+  shell-file: release.nix
```
ksaric: The convention from `cardano-chain` is to use a `scripts` folder and from `cardano-sl` is to use a `pkgs` folder.
I agree that there is no need to pollute the root directory so we can leave it there.

ksaric:
See https://github.com/input-output-hk/cardano-shell/pull/5

release.nix
robcohen: Is there a source of truth for current versioning information that I can refer to in the future? Perhaps something in the cardano-sl repo?

release.nix
robcohen: I believe release.nix is required by Hydra.

scripts/nix/fetchNixpkgs.nix
robcohen: Appears to be used by cardano-sl, not sure how much of their nix configs we should use.

scripts/nix/nixpkgs-src.json
robcohen: sha256unpacked is needed by fetchNixpkgs, but if we are getting rid of that then it makes sense to not add this.

stack.yaml
robcohen: Agree, lets minimize root folder nix scripts. We should be able to get it down to release.nix and shell.nix.

PR #5, opened 2018-10-24
#3 Add simple CI.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/3

Adding simple BuildKite CI.


ksaric:
Am I missing something or are they really all different?
- `./cabal list --installed`
- `stack ls dependencies`
- `nix-shell release.nix --run 'ghc-pkg list' | grep universum`

ksaric:
There are three situations where I really have no clue what's happening.

- This fails for me - `haskell.lib.buildStackProject { ... }`, changing the shell to the one we had before simply doesn't  work. 
- Try finding `universum` with `nix-shell release.nix --run 'ghc-pkg list'`.
- Try switching to Stack, change the Universum to 1.3.0 in the `release.nix` and rebuild. After that, try `stack ls dependencies`. What is the version? 1.2.0! But it uses Nix, or at least it should.

ksaric:
Requested a review from DevOps since I don't know if this is sensible.

robcohen:
CI works correctly, removed old nix files.

PR #10, opened 2018-11-01
Buildkite pipeline with stack rebuild

rvl:
This uses the common `snapshot.yaml` from cardano-prelude, which we don't have yet for nix builds.


rvl:
Thanks. The stackage/nix build is in #11.

PR #13, opened 2018-11-06
Fix badge to reflect only develop

robcohen:
This PR fixes the issue with the build badge checking all branches instead of only develop.


robcohen:
Added a table and both badges as requested.

ksaric:
@robcohen I don't see the badges in the UI and I thought that you meant badges for `develop` and for `master`.

robcohen:
Now shows "develop" and "master". Master doesn't work I believe because it is a new branch and nothing has been pushed yet, but I assume it will work once a CI pipeline for it is started after a push.

Issue #15, opened 2018-11-08
Explore initial design of the `cardano-shell` architecture


ksaric:
Play around with the initial design until we have an example we can share with the rest of the teams.
As discussed in the meeting, some initial ideas can be found here:
- https://github.com/input-output-hk/cardano-shell/tree/ksaric/plugin-idea
- https://github.com/input-output-hk/cardano-shell/tree/ksaric/features-work (mostly copied from existing work by @HirotoShioi - https://gist.github.com/HirotoShioi/85cf32383c7c846da8f77147c93abeb9#file-serverframeworkv2-hs-L127)

PR #17, opened 2018-11-21
[GH-15] Explore initial design of the `cardano-shell` architecture

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/15

An improved version as suggested. Output when running(notice the timing in the first two features) :
```
Starting up logging!
Starting up networking!
Starting up networkingCardanoFeature!
Starting up loggingCardanoFeature!
Running node/wallet/whatever!
Running node/wallet/whatever!
Running node/wallet/whatever!
Running node/wallet/whatever!
Running node/wallet/whatever!
Exception occured!
Shutting down networkingCardanoFeature!
Shutting down networking feature!
Shutting down loggingCardanoFeature!
Shutting down logging feature!
```

### DIFF
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -1,0 +1,69 @@
+module Types where
+
+import Cardano.Prelude
+
+--import qualified Katip as Katip
+import qualified System.Metrics as Ekg
+
+-- | Cardano configuration
+data CardanoConfiguration = CardanoConfiguration
+    { scfgLogPath   :: !FilePath
+    , scfgDBPath    :: !FilePath
+    , scfgSomeParam :: !Int
+    }
+
+-- | The common runtime environment for all features in the server.
+-- All features have access to this environment.
+data CardanoEnvironment = CardanoEnvironment
+    { serverLogEnv      :: Text --Katip.LogEnv
+    , serverEkgStore    :: Ekg.Store
+     -- ...
+    }
+
+-- | Initalise 'ServerEnv'
+initializeCardanoEnvironment :: IO CardanoEnvironment
+initializeCardanoEnvironment = do
+  --  logenv   <- Katip.initLogEnv (...) (...)
+    ekgStore <- Ekg.newStore
+    return CardanoEnvironment {serverLogEnv = "To implement", serverEkgStore = ekgStore}
+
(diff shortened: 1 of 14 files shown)


src/Types.hs
```diff
@@ -0,0 +1,75 @@
...
+
+
+-- | The feature types we can have in the project.
+data FeatureType
+    = LoggingMonitoringFeature
+    | NetworkingFeature
+    | BlockchainFeature
+    | LedgerFeature
+    | WalletFeature
+    deriving (Eq, Show)
```
dcoutts: Do we need this or is it just for information? If it's just for info I'd use a string instead. The framework should not generally have to know the complete list of features.

src/Types.hs
```diff
@@ -0,0 +1,75 @@
...
+-- | Cardano feature initialization.
+-- We are saying "you have the responsibility to make sure you use the right context!".
+data CardanoFeatureInit dependency configuration layer = CardanoFeatureInit
+    { featureType                   :: !FeatureType
+    -- ^ The type of the feature that we use.
+    , featureInit                   :: CardanoEnvironment -> dependency -> CardanoConfiguration -> configuration -> IO layer
+    -- ^ Again, we are not sure how is the user going to run the actual feature,
+    -- so we provide him with the most flexible/powerful context we have, @IO@.
+    -- Notice the arrangement of the parameters - specific, general, specific, general, result.
+    , featureCleanup                :: layer -> IO ()
```
dcoutts: Do we need both a `featureCleanup` and `featureShutdown`?

src/Types.hs
ksaric: Just type safe info. Yes, there is a tradeoff here, but we will have a point where we will have to define all the features so it might be a good idea to have them (type)checked.

src/Types.hs
ksaric: `featureShutdown` is for explicitly shutting down the feature (when the developers want to do that) and `featureCleanup` is when the developers want to clean something up before the feature is shut down - for example, saving some logs or closing files or something similar.

`featureShutdown` is under the control of the developers, while `featureCleanup` is a bracket cleanup call, basically.

Issue #14, opened 2018-11-08
Reuse existing DHall configuration 


ksaric:
As mentioned here - https://github.com/input-output-hk/cardano-shell/issues/1#issuecomment-433957215, we can reuse most of the configuration from the already existing DHall configuration files in Daedalus - https://github.com/input-output-hk/daedalus/tree/develop/installers/dhall

This issue is covering the migration of the configuration files and their respective parsers to the `cardano-shell` project where we can reuse most of the existing work.

Of course, we are not interested in Daedalus configuration options for now, but we can most all of them to see to what extent can they be transferred to the `cardno-shell` repo.

In translation - copy as much configuration as possible that is related to the _node_, for now, leave out Daedalus configuration options out.

HirotoShioi:
Here's the PR:
https://github.com/input-output-hk/cardano-shell/pull/11

HirotoShioi:
Here's the PR!
https://github.com/input-output-hk/cardano-shell/pull/18

PR #18, opened 2018-11-22
[GH-14] Reuse existing Dhall configuration

HirotoShioi:
[#14](https://github.com/input-output-hk/cardano-shell/issues/14)

Still in progress


dhall/mainnet.dhall
```diff
@@ -0,0 +1,9 @@
+{ name         = "mainnet"
+, keyPrefix    = "mainnet_wallet"
+, relays       = "relays.cardano-mainnet.iohk.io"
+, updateServer = "https://update-cardano-mainnet.iohk.io"
+, reportServer = "http://report-server.cardano-mainnet.iohk.io:8080"
+, installDirectorySuffix = ""
+, macPackageSuffix       = ""
+, walletPort             = 8090
```
ksaric: For example, `walletPort` belongs in a wallet specific configuration, as does `reportServer` and possibly some others. My idea is to have the _base configuration_ and the have the _per feature configuration_ (so for wallet, we can have _walletConfig.dhall_).

HirotoShioi:
How to run an parser:
`mkLauncher <OS> <Cluster>`
OS could be one of the following:
  Linux64
  MacOS64
  Win64

Cluster:
 Mainnet
 Staging
 Testnet
 Demo

HirotoShioi:
Note: I don't think there's an way to implement a test case on serialization roundtrip since Dhall does not export encode function. You can decode Dhall, but, but cannot encode them.

src/Configuration/Types.hs
```diff
@@ -0,0 +1,383 @@
...
+import           GHC.Generics (from, to)
+import           Test.QuickCheck (Arbitrary (..), Gen, arbitraryASCIIChar,
+                                  choose, elements, listOf, listOf1)
+
+-- | Operating system
+data OS
+    = Linux64
+    | Macos64
+    | Win64
+    deriving (Bounded, Enum, Eq, Read, Show)
```
ksaric: Maybe you don't need `Show` if you have specific functions for rendering these into `Text`. Do you really need anything more than `Eq` here?

src/Configuration/Types.hs
```diff
@@ -0,0 +1,383 @@
...
+
+instance Interpret OSConfig where
+    autoWith opt = interpretAutoWithOption $ opt {fieldModifier = lowerHead . T.drop 2}
+
+instance Inject OSConfig where
+    injectWith opt = injectAutoWithOption $ opt {fieldModifier = lowerHead . T.drop 2}
+
+instance Arbitrary OSConfig where
+    arbitrary = do
+        osName              <- elements ["windows64", "macos64", "linux64"]
```
ksaric: Here you could have used the `Enum` instance to generate the `OS` via `renderOS`.
Otherwise, the person adding the OS (which might be possible) might forget about this instance here.

src/Configuration/Types.hs
```diff
@@ -0,0 +1,383 @@
...
+    { osName              :: !Text
+    , osConfigurationYaml :: !Text
+    , osInstallDirectory  :: !Text
+    , osX509ToolPath      :: !Text
+    , osNodeArgs          :: !NodeArgs
+    , osPass              :: !Pass
+    } deriving (Eq, Generic, Show)
+
+instance Interpret OSConfig where
+    autoWith opt = interpretAutoWithOption $ opt {fieldModifier = lowerHead . T.drop 2}
```
ksaric: I really don't like these `T.drop` and transformations. It could be very error-prone and detected only on runtime (yes, tests might catch this, but it would be better if we don't have to do that).
We could simply keep things simple and name them _same as in the configuration file_ and be done with it.
The cost is a couple of letters as a prefix, and I consider that to be acceptable.

src/Configuration/Types.hs
```diff
@@ -0,0 +1,383 @@
...
+        naLogsPrefix       <- genSafeText
+        naTopology         <- genSafeText
+        naUpdateLatestPath <- genSafeText
+        naWalletDBPath     <- genSafeText
+        naTlsPath          <- genSafeText
+
+        pure NodeArgs{..}
+
+-- | Paths
+data Pass = Pass
```
ksaric: I really don't understand why it's called `Pass`. Maybe `Paths` was the intention?

src/Configuration/Types.hs
```diff
@@ -0,0 +1,383 @@
...
+        pUpdateArchive       <- maybeOf genSafeText
+        pUpdateWindowsRunner <- maybeOf genSafeText
+        pLauncherLogsPrefix  <- genSafeText
+
+        pure Pass{..}
+-- | Launcher configuration
+data LauncherConfig = LauncherConfig
+    { lcfgFilePath    :: !Text
+    , lcfgKey         :: !Text
+    , lcfgSystemStart :: !(Maybe Natural)
```
ksaric: Oh? Is there no way to convert the `Natural` to a plain `Integer` here? I understand you had issues with the types, but still looks kind of funny - not really sure if the author intended to tie our hands using only `Natural` and not something idiomatic like `Integer`.

src/Configuration/Types.hs
```diff
@@ -0,0 +1,383 @@
+{-# LANGUAGE DeriveGeneric     #-}
+{-# LANGUAGE FlexibleContexts  #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards   #-}
```
ksaric: Maybe drop this and make everything very explicit? No need to have people wonder how the structure got filled with values, and this is not really good if we want to use thing configuration elsewhere. Erik had a nice talk about this.

src/Configuration/Types.hs
```diff
@@ -0,0 +1,383 @@
...
+
+-- | Launcher configuration
+data Launcher = Launcher
+    { lConfiguration     :: !LauncherConfig
+    , lNodeDbPath        :: !Text
+    , lNodeLogConfig     :: !Text
+    , lUpdaterPath       :: !Text
+    , lUpdaterArgs       :: ![Text]
+    , lUpdateArchive     :: !(Maybe Text)
+    , lReportServer      :: !Text
```
ksaric: If we intend to use this throughout the codebase, we probably need to add `newtype` wrappers. 
In some other PR would be fine, just commenting.

stack.yaml
```diff
@@ -73,5 +73,6 @@ extra-deps:
 # compiler-check: newer-minor
 
 nix:
+  enable: true
```
ksaric: I presume we need this because of `Dhall`?

test/DhallConfigSpec.hs
```diff
@@ -0,0 +1,50 @@
...
+import           Dhall.Core (pretty)
+import           Test.Hspec (Spec, describe)
+import           Test.Hspec.QuickCheck (modifyMaxSuccess, prop)
+import           Test.QuickCheck (Property)
+import           Test.QuickCheck.Monadic (assert, monadicIO, run)
+
+import           Configuration.Types (ClusterConfig, InstallerConfig, Launcher,
+                                      NodeArgs, Pass, OSConfig)
+
+dhallConfigSpec :: Spec
```
ksaric: Very nice, Hiroto. Great job.

src/Configuration/Types.hs
HirotoShioi: Well, there's two ways to do this:
1. Modify the content of dhall files so it can be parsed as `Integer`. So I'll need to modify something like this:

```
-- Add plus
walletPort = +8090
````

2. Bump the dhall's version to the latest one.
The current version of dhall that the shell use (1.5.1) does not have much freedom when it comes to defining an instance of `Inject` which is a instance needed to encode given data type into dhall format.
The latest version of dhall (1.9.1) on the other hand seems to have this feature (http://hackage.haskell.org/package/dhall-1.19.1/docs/Dhall.html#v:-62--42--60-).
To do so, I'd need to either 
1. Add an lot of `extra-deps` into the shell project
2. Use nix package system to use certain version of `dhall` package

Comment (HirotoShioi) [reply src/Configuration/Types.hs]: For now, I'll apply step 1 since it's the most safest and easiest one.

File: src/Configuration/Types.hs (hunk 4 above)
Comment (HirotoShioi) [reply src/Configuration/Types.hs]: @ksaric In that case, what should I do with the fields that has exact same name? For example `relays`
https://github.com/input-output-hk/cardano-shell/blob/hiroto/reuse-dhall/src/Configuration/Types.hs#L67
https://github.com/input-output-hk/cardano-shell/blob/hiroto/reuse-dhall/src/Configuration/Types.hs#L282

`name`
https://github.com/input-output-hk/cardano-shell/blob/hiroto/reuse-dhall/src/Configuration/Types.hs#L107
https://github.com/input-output-hk/cardano-shell/blob/hiroto/reuse-dhall/src/Configuration/Types.hs#L65

File: stack.yaml (hunk 9 above)
Comment (HirotoShioi) [reply stack.yaml]: Yes!

File: src/Configuration/Types.hs (hunk 5 above)
Comment (HirotoShioi) [reply src/Configuration/Types.hs]: This is one of the field that was on
https://github.com/input-output-hk/daedalus/blob/develop/installers/dhall/os.type#L14
so I've just copied it. I'll check with devops about it

Comment (HirotoShioi) [reply src/Configuration/Types.hs]: I've checked with devops and told me `pass` is just an name of variable so I'll change chang to `Param`!

File: src/Configuration/Types.hs (hunk 4 above)
Comment (ksaric) [reply src/Configuration/Types.hs]: No, I meant modify the configuration files names to reflect the names in Haskell.

File: src/Configuration/Types.hs (hunk 6 above)
Comment (ksaric) [reply src/Configuration/Types.hs]: Sounds like the most straightforward solution.

File: src/Configuration/Types.hs (hunk 4 above)
Comment (HirotoShioi) [reply src/Configuration/Types.hs]: Ahh okay

**Comment by HirotoShioi** [comment]
@ksaric I've applied changes on the #18 (comment)

File: docs/Dhall.md [hunk 11]
```diff
@@ -0,0 +1,39 @@
...
+This means I had to switch some of the types (e.g. WalletPort) from Integer to Natural
+
+## Defining an instance of Interpret with defaultInterpretOptions causes infinite loop
+
+In order to parse Dhall file, you'll need to define an typeclasss instance of `Interpret`, similiar [aeson](http://hackage.haskell.org/package/aeson-1.4.1.0/docs/Data-Aeson.html).
+
+
+There's is an way to easier way to define an instance of it using [defaultInterpretOptions](http://hackage.haskell.org/package/dhall-1.15.1/docs/Dhall.html#v:defaultInterpretOptions).
+
+But for some reason, this causes infinite loop. Currently looking into it.
```
ksaric: Maybe you don't need this sentence anymore? The rest is fine.

Issue #19, opened 2018-11-26
Improve initialization and run for the `cardano-shell`


ksaric:
As Duncan mentioned, the startup could be improved - https://github.com/input-output-hk/cardano-shell/pull/17#pullrequestreview-178272699 and we can have two strategies:

1. dev/debug mode strategy: shut down all features and terminate with the exception.
2. production mode strategy: shut down all features and restart them all from scratch.

For example, run them all with `async`, wait for any exception, then call `cancel` on them in the reverse list order.

PR #20, opened 2018-11-28
[GH-19] Improve initialization and run for the cardano-shell.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/19

Improved initialization and run, added an option for restart.


src/Cardano/Shell/Lib.hs
```diff
@@ -0,0 +1,102 @@
...
+-- General exceptions
+--------------------------------------------------------------------------------
+
+data GeneralException
+    = UnknownFailureException -- the "catch-all"
+    | MissingResourceException
+    | FileNotFoundException
+    deriving (Eq, Show)
+
+instance Exception GeneralException
```
erikd: Do we really want to use exceptions? Especially since the constructors `MissingResourceException` and `FileNotFoundException` do not sound much like exceptions. They are flow control.

ksaric: At this point, they are just examples of some things that can go wrong that can't be handled from the actual module/feature. 

> They are flow control.

Well, when a module/feature doesn't find a file that it requires for running, what is the result of that control flow? I presume termination - so the point here is to have a top-level information about that termination and why it happened and restart it.

In any case, we don't have actual examples of how we could use this, it's just my idea of what would be a nice general way to inform the top level application of what went wrong.

src/Cardano/Shell/Lib.hs
```diff
@@ -0,0 +1,102 @@
...
+
+    -- When we reach this point, we cancel all the features.
+    _ <- mapM cancel (reverse asyncCardanoFeatures)
+
+    -- Then we cleanup all the features if we need to do so.
+    _ <- mapM featureShutdown (reverse cardanoFeatures)
+
+    -- And we are done! Or are we? A simple idea is to restart the application if it's
+    -- in production.
+    if applicationProductionMode applicationEnvironment
```
erikd: Would use `when` there.

Issue #8, opened 2018-10-26
Migrate configuration.yaml to Haskell


ksaric:
The configuration file (https://github.com/input-output-hk/cardano-sl/blob/develop/lib/configuration.yaml) is too large and contains things that are really not configurations.
They would be mostly constants.

The task is to create a Haskell module on top of https://github.com/input-output-hk/cardano-shell/pull/7 which contains all this information _in this repository_. 
There is a common structure, so let's use a common structure.

erikd:
The thing to note is that I suspect some of the item in that configuration object are not even needed in the post OBFT version. Anything to do with the VSS (verifiable secret sharing), LRC, SSC can probably be dropped because they are not relevant for OBFT and probably different for the new Ouroboros protocols.

robcohen:
Created a bunch of data types for the configuration.

robcohen:
Compiled a description of the variables in one place for quick ref: https://github.com/input-output-hk/cardano-shell/blob/robcohen/GH-8-configuration-yaml-migration/configuration.md

Still attempting to determine a metric to find which variables should be hardcoded vs not.

robcohen:
I kindly request that my commits be reviewed when possible. There is a CardanoConfig for both dev and mainnet.

PR #33, opened 2019-01-02
[GH-8] configuration yaml migration

ksaric:


app/Cardano/Shell/Features/Logging.hs
```diff
@@ -1,5 +1,6 @@
-{-# LANGUAGE DeriveGeneric #-}
-{-# LANGUAGE Rank2Types    #-}
+{-# LANGUAGE DeriveGeneric     #-}
+{-# LANGUAGE OverloadedStrings #-}
```
ksaric: This is included in the cabal as the default extension. Or it should be included.

src/Cardano/Shell/Types.hs
```diff
@@ -1,6 +1,18 @@
-{-# LANGUAGE Rank2Types #-}
+{-# LANGUAGE OverloadedStrings #-}
```
ksaric: Again, this is extra, or it should be extra.

src/Cardano/Shell/Types.hs
```diff
@@ -89,3 +87,468 @@ data CardanoFeature = CardanoFeature
     -- ^ What we call when we shut down the feature.
     }
 
+--------------------------------------------------------------------------------
+-- Cardano Mainnet Configuration
+--------------------------------------------------------------------------------
+
+mainnetConfiguration :: CardanoConfiguration
```
ksaric: Maybe move this into another module? It's not bad like this, but it might get tricky later on...

cardano-shell.cabal
```diff
@@ -51,6 +52,8 @@ library
   default-language:    Haskell2010
   default-extensions:  NoImplicitPrelude
                        OverloadedStrings
+                       DeriveGeneric
```
ksaric: I don't really think we need these in default-extensions.
People should have visibility when looking at the source code and adding this here removes some of that visibility.

src/Cardano/Shell/Presets.hs
```diff
@@ -0,0 +1,316 @@
+{-# LANGUAGE RecordWildCards #-}
```
ksaric: Ah, a bit bittersweet but if it's local to this module only, fine.

src/Cardano/Shell/Presets.hs
```diff
@@ -0,0 +1,316 @@
...
+-- Cardano Mainnet Configuration
+--------------------------------------------------------------------------------
+
+mainnetConfiguration :: CardanoConfiguration
+mainnetConfiguration = CardanoConfiguration
+                       { ccLogPath  = "./logs/"
+                       , ccLogConfig = "./log-config.yaml"
+                       , ccDBPath   = "./db/"
+                       , ccApplicationLockFile = ""
+                       , ccCore       = Core { coGenesis =
```
ksaric: Can you format this better? It's still all over the place.

src/Cardano/Shell/Types.hs
```diff
@@ -1,6 +1,38 @@
-{-# LANGUAGE Rank2Types #-}
-
-module Cardano.Shell.Types where
+module Cardano.Shell.Types
```
ksaric: Please do move these types in a separate module, as I mentioned.
We don't want to pollute the top-level namespace with things that are not its responsibility.

src/Cardano/Shell/Types.hs
```diff
@@ -53,9 +71,9 @@ initializeCardanoEnvironment = do
         , ceEkgStore    = ekgStore
         }
 
-loadCardanoConfiguration :: IO CardanoConfiguration
-loadCardanoConfiguration = do
-    pure $ CardanoConfiguration mempty mempty mempty mempty
+loadCardanoConfiguration :: CardanoConfiguration -> IO CardanoConfiguration
+loadCardanoConfiguration configuration = do
```
ksaric: This seems obsolete then unless you pass all the things that `CardanoConfiguration` requires.

src/Cardano/Shell/Types.hs
robcohen: Oh I thought that was just the mainnetConfiguration and devConfiguration.
Should I move everything below "-- Cardano Configuration Data Structures" into ConfigurationTypes.hs ?

ksaric: You can try the folder/package `Constants` - so `Constants.Types`.

robcohen: Changes implemented.

src/Cardano/Shell/Presets.hs
robcohen: Not sure how to properly style this. Stylish haskell isn't fixing it for me automatically, and I haven't had nested records this deep before. Any recommendations?

src/Cardano/Shell/Presets.hs
robcohen: I can probably take it out, just putting it on hold until I fix the other issues.

src/Cardano/Shell/Types.hs
robcohen: Seems like I'll need to write a lot of boilerplate  `instance Monoid` for the nested data types. Perhaps https://hackage.haskell.org/package/generic-monoid is a better option?

ksaric: You could simply derive it in the deriving clause?

app/Cardano/Shell/Features/Logging.hs
```diff
@@ -43,7 +44,11 @@ data LoggingParameters = LoggingParameters
 -- the functions effects and constraining the user (programmer) of those function to use specific effects in them.
 -- https://github.com/input-output-hk/cardano-sl/blob/develop/util/src/Pos/Util/Log/LogSafe.hs
 data LoggingLayer = LoggingLayer
+<<<<<<< HEAD
```
ksaric: That is left from the merge conflicts.

src/Cardano/Shell/Constants/Types.hs
```diff
@@ -0,0 +1,275 @@
+module Cardano.Shell.Constants.Types
```
ksaric: Yep, nice.

src/Cardano/Shell/Presets.hs
ksaric: Manual formatting is what I tend to do?

src/Cardano/Shell/Types.hs
```diff
@@ -55,9 +66,20 @@ initializeCardanoEnvironment = do
 loadEmptyCardanoConfiguration :: IO CardanoConfiguration
```
ksaric: I think this is in conflict with the newest changes.

cardano-shell.cabal
robcohen: Removed.

app/Cardano/Shell/Features/Logging.hs
robcohen: I should have caught this. Corrected.

src/Cardano/Shell/Types.hs
robcohen: I believe Generics are required, no? https://stackoverflow.com/questions/49517453/defining-a-monoid-instance-for-a-record-type

src/Cardano/Shell/Presets.hs
robcohen: I tried another formatting structure. Any recommended changes @ksaric ?

src/Cardano/Shell/Types.hs
robcohen: Still getting warning about implicit <> for Semigroup, but it compiles at least.

src/Cardano/Shell/Presets.hs
robcohen: Removed.

src/Cardano/Shell/Constants/Types.hs
```diff
@@ -0,0 +1,276 @@
...
+    , ccNTP                 :: !NTP
+    , ccUpdate              :: !Update
+    , ccTXP                 :: !TXP
+    , ccSSC                 :: !SSC
+    , ccDLG                 :: !DLG
+    , ccBlock               :: !Block
+    , ccNode                :: !Node
+    , ccTLS                 :: !TLS
+    , ccWallet              :: !Wallet
+    } deriving (Eq, Show, Monoid, Semigroup)
```
ksaric: Do you need the `Monoid` and `Semigroup`?

.gitignore
```diff
@@ -60,6 +60,12 @@ out/
 # Atom
 .haskell-ghc-mod.json
 
+# VSCode
+cardano-shell.code-workspace
+
+# GHCID Output
+ghcid.txt
```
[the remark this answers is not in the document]
robcohen: ghcid is just for debugging purposes. Also the plus is just github.

src/Cardano/Shell/Constants/Types.hs
robcohen: Removed, thank you for pointing that out.

cardano-shell.cabal
```diff
@@ -51,6 +53,7 @@ library
   default-language:    Haskell2010
   default-extensions:  NoImplicitPrelude
                        OverloadedStrings
+                       Rank2Types
```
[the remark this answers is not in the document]
robcohen: Apparently this is required by some code:
```cardano-shell/src/Cardano/Shell/Types.h
77 |     , featureStart    :: forall m. (MonadIO m, MonadConc m)```

Issue #39, opened 2019-01-15
Add tests for testing arbitrary CardanoFeature concurrency issues.


ksaric:
After #27 the actual concurrency tests were made obsolete since we moved the actual specific CardanoFeature(s) to the executable.
We need to implement tests that will check if arbitrary CardanoFeatures will work correctly and what are the conditions in which they will not work correctly.

PR #42, opened 2019-01-23
[GH-39] Add tests for testing arbitrary CardanoFeature concurrency issues.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/39

Added tests for testing the correctness of the arbitrary modules.


test/Spec.hs
```diff
@@ -56,7 +58,82 @@ validConcurrencySpec = do
...
+        genSafeText = strConv Lenient <$> listOf1 arbitraryASCIIChar
+
+-- | The function that generates a single command in @IO@.
+generateSingleCommand :: (MonadIO m) => m ()
+generateSingleCommand = do
+    generatedCommand <- liftIO $ generate concurrencyDSLGenerator
+    interepretConcurrencyDSL generatedCommand
+
+-- | The function that creates a list of commands in @IO@.
+generateListOfCommands :: (MonadIO m) => m ()
```
HirotoShioi: This function as well as `concurrencyDSLSequenceGenerator` is not being used. Are we planning to use them on future?

test/Spec.hs
```diff
@@ -56,7 +58,82 @@ validConcurrencySpec = do
...
+-- | The DSL generator. It uses specific frequencies so it a bit biased,
+-- but I think it's showing a more realistic generation like this.
+concurrencyDSLGenerator :: Gen CommandDSL
+concurrencyDSLGenerator = frequency
+    [ (5, SequentialCmd <$> generateMillisecondDelay)
+    , (3, ParallelCmd   <$> generateMillisecondDelay)
+    , (2, AsyncCmd      <$> generateMillisecondDelay)
+    ]
+ where
+    -- | A delay from one millisecond to one whole day.
```
HirotoShioi: one to ten millisecond?

ksaric:
@HirotoShioi Ready for review.

test/Spec.hs
robcohen: On this note, what would be a good interval for testing various features? It seems like it should be grouped into separate sets for testing (1-10 ms, 1000-10000, 10000-100000 for example). A whole day might be problematic for testing purposes as well, as running the tests may take too long? @ksaric what are your thoughts?

ksaric: @robcohen Since this test is actually being checked by DejaFu (or the intention behind is that we test it using it), my understanding is that a large range/interval covering our domain actual values would be ideal, but it turns out that if you increase the delay, the tests slow down.
Ideally, I would like to sit down and write something like this myself and tell you "it doesn't matter", but these things are really complicated and take up a ton of time.
To sum up:
- ideally, you should use the actual domain values (if you have up to one day then one millisecond up to one day)
- in this case, using larger delay slows the computation significantly. 

The case where the slowdown appears in cases when there are too many modules is understandable (combinatorial explosion), but in these cases with delay, it's not really clear to me why this is happening and it could be my lack of understanding how the underlying library works.

PR #48, opened 2019-01-30
Configuration parsing of logging layer.

365andreas:
Pass a configuration file to logging feature in order to be initialized.
Added also some demo calls of logging functions from Logging Layer in `app/Main.hs`


app/Main.hs
```diff
@@ -12,28 +13,34 @@ main :: IO ()
...
             pure ()
 
-    let cardanoApplication :: CardanoApplication
-        cardanoApplication = CardanoApplication application
+    let cardanoApplication :: LoggingLayer -> CardanoApplication
+        cardanoApplication = CardanoApplication . application
 
     -- Here we initialize all the features.
-    cardanoFeatures <- initializeAllFeatures cardanoConfiguration cardanoEnvironment
+    (cardanoFeatures, loggingLayer) <- initializeAllFeatures cardanoConfiguration cardanoEnvironment
```
ksaric: Right, with the addition that here we return the **top layer** that we need to run.

app/Cardano/Shell/Features/Logging.hs
```diff
@@ -62,9 +61,8 @@ createLoggingFeature cardanoEnvironment cardanoConfiguration = do
     -- we parse any additional configuration if there is any
     -- We don't know where the user wants to fetch the additional configuration from, it could be from
     -- the filesystem, so we give him the most flexible/powerful context, @IO@.
-    loggingConfiguration    <-  pure $ LoggingParameters
-                                    (ccLogConfig cardanoConfiguration)
-                                    (ccLogPath cardanoConfiguration)
+    loggingConfiguration    <-  pure $ LoggingParameters $
```
ksaric: The idea is that you _load your configuration_ here, not pass the name of the configuration that needs to be loaded on a function that then loads the configuration. Reduce indirection.

app/Cardano/Shell/Features/Logging.hs
```diff
@@ -27,8 +27,7 @@ import           Cardano.Shell.Types (CardanoConfiguration, CardanoEnvironment,
 --------------------------------
 
 data LoggingParameters = LoggingParameters
-    { configFp :: FilePath
-    , prefixFp :: FilePath
+    { lpFilePath :: !FilePath
```
ksaric: The idea is that here you have actual parameters from the logging module, and when you load them in the initialization, you pass them on.

CodiePP:
@ksaric : can this PR be merged?

Issue #57, opened 2019-02-11
Add specification for update system


ksaric:
Adding a specification for the update system. The current situation is that the update system is pretty complicated. We need to try to find a way to make it simpler and writing a simple spec up front should be a nice way to see what is important and what is not.

PR #60, opened 2019-03-01
[GH-57] Add specification for update system

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/57


src/Cardano/Shell/Update/Types.hs
```diff
@@ -0,0 +1,114 @@
...
+    deriving (Eq, Show)
+
+data InstallerVersion = InstallerVersion InstallerHash
+    deriving (Eq, Ord, Show)
+
+data InstallerHash = InstallerHash !Text
+    deriving (Eq, Ord, Show)
+
+-- | The wrapper is used since this is an important distinction.
+newtype LatestInstallerVersion = LatestInstallerVersion
```
HirotoShioi: You should wrap it with newtype instead of data.
Same goes to Epoch, Slot, BlockHash, InstallerVersion.

src/Cardano/Shell/Update/Types.hs
```diff
@@ -0,0 +1,114 @@
...
+    [ (1 , Just <$> genInstallerVersion)
+    , (10, pure Nothing)
+    ]
+
+-- | Here we generate the blockchain that contains the installer
+genBlockchainWithInstaller :: Gen Blockchain
+genBlockchainWithInstaller = do
+    numOfEpochs <- choose (1, 100)
+
+    let availableEpochs = map Epoch [1..numOfEpochs]
```
HirotoShioi: You can use `coerce` if `Epoch` is defined with `newtype`

src/Cardano/Shell/Update/Types.hs
```diff
@@ -0,0 +1,114 @@
...
+-- Simply iterate the blockchain and if there exists an @InstallerVersion@
+-- append it to the existing list.
+fetchUpdatesFromBlockchain :: Blockchain -> [InstallerVersion]
+fetchUpdatesFromBlockchain (Blockchain blockchain) =
+    let foldrInstallerVersions installerVersions (_epoch, slots) = installerVersions <> filter isJust (map slotContainsInstaller slots)
+    in  catMaybes (foldl foldrInstallerVersions mempty (M.toList blockchain))
+
+-- | Fetch latest installer version, or latest update from the blockchain.
+latestUpdateFromBlockchain :: Blockchain -> Maybe LatestInstallerVersion
+latestUpdateFromBlockchain = fmap LatestInstallerVersion . head . reverse . fetchUpdatesFromBlockchain
```
HirotoShioi: `last` instead of `head . reverse`

HirotoShioi:
Also in the figure7 on the spec, it's not clear for me how `cardano-launcher` checking the presence of installer file affects the rest of the process.

src/Cardano/Shell/Update/Types.hs
ksaric: Usually. Since this is not a part of any actual production code, it really isn't that critical.
It's been left this way, since it's easier to change in this form. Using a newtype wrapper removes the ability to add additional constructors...

src/Cardano/Shell/Update/Types.hs
ksaric: Do NOT use coerce. It makes things unclear and less explicit.

src/Cardano/Shell/Update/Types.hs
ksaric: Meh. It's an ok idea, but I don't really see much benefit.

ksaric:
Fixed build errors, fix found here - https://github.com/input-output-hk/cardano-prelude/commit/ff5fd5f33849be8a826506c34e5b0278f267f804

PR #68, opened 2019-03-12
Resolve technical debt

HirotoShioi:
- [x] https://github.com/input-output-hk/cardano-shell/blob/develop/src/Cardano/Shell/NodeIPC/Lib.hs#L173
- [x] https://github.com/input-output-hk/cardano-shell/blob/develop/src/Cardano/Shell/NodeIPC/Lib.hs#L198
- [x] https://github.com/input-output-hk/cardano-shell/blob/develop/src/Cardano/Shell/NodeIPC/Lib.hs#L188
- [x] https://github.com/input-output-hk/cardano-shell/blob/develop/src/Cardano/Shell/NodeIPC/Example.hs#L82
- [x] https://github.com/input-output-hk/cardano-shell/blob/develop/src/Cardano/Shell/NodeIPC/Message.hs#L73
- [x] Clean up resources properly when exception occurs or process is finished.

### DIFF
--- a/src/Cardano/Shell/NodeIPC/Example.hs
+++ b/src/Cardano/Shell/NodeIPC/Example.hs
@@ -79,15 +79,16 @@ exampleWithProcess = do
     (clientReadHandle, clientWriteHandle) <- getReadWriteHandles
 
     -- Create a child process that acts as an server
-    -- Can't apply bracket pattern therefore vulnerable to async exception
-    -- (e.g crash the program in the middle of process)
-    _ <- forkProcess $ do
-        (serverReadHandle, serverWriteHandle) <- getReadWriteHandles
-        -- Send message to server
-        sendMessage serverWriteHandle Ping
-        let nodePort = Port 8090
-        startNodeJsIPC serverReadHandle clientWriteHandle nodePort
-        exitImmediately ExitSuccess
+    _ <- finally  
+            (forkProcess $ do
+                (serverReadHandle, serverWriteHandle) <- getReadWriteHandles
+                -- Send message to server
+                sendMessage serverWriteHandle Ping
+                let nodePort = Port 8090
+                startNodeJsIPC serverReadHandle clientWriteHandle nodePort
+                exitImmediately ExitSuccess
+            )
+            (return ())
 
     let readClientMessage :: IO MsgOut
         readClientMessage = readMessage clientReadHandle
--- a/src/Cardano/Shell/NodeIPC/Example.hs
+++ b/src/Cardano/Shell/NodeIPC/Example.hs
@@ -29,7 +29,7 @@ import           System.Posix.Process (exitImmediately, forkProcess)
 import           System.Process (createPipe)
 
 import           Cardano.Shell.NodeIPC.Lib (MsgIn (..), MsgOut (..), Port (..),
-                                            startNodeJsIPC)
+                                            startIPC)
 import           Cardano.Shell.NodeIPC.Message (ReadHandle (..),
                                                 WriteHandle (..), readMessage,
                                                 sendMessage)
@@ -57,7 +57,7 @@ exampleWithFD = do
     (serverReadHandle, serverWriteHandle) <- getReadWriteHandles
 
     let nodePort = Port 8090
-    asyncServer <- async $ startNodeJsIPC serverReadHandle clientWriteHandle nodePort
+    asyncServer <- async $ startIPC serverReadHandle clientWriteHandle nodePort
     link asyncServer
 
     -- Use these functions so you don't pass the wrong handle by mistake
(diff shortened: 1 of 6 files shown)


src/Cardano/Shell/NodeIPC/Lib.hs
```diff
@@ -139,61 +161,112 @@ getIPCHandle = do
...
+        when (not result) $ throwM exception
 
 --------------------------------------------------------------------------------
--- placeholder
+-- Placeholder
 --------------------------------------------------------------------------------
 
+-- | Use this until we find suitable logging library
 logError :: Text -> IO ()
 logError _ = return ()
```
ksaric: What about simply outputting this message on `stdOut` or `stdErr` for now?

src/Cardano/Shell/NodeIPC/Lib.hs
```diff
@@ -139,61 +161,112 @@ getIPCHandle = do
...
         case msgIn of
-            QueryPort -> send (ReplyPort port) >> shutdown
-            Ping      -> do
-                send Pong
-                shutdown
+            QueryPort          -> send (ReplyPort port)
+            Ping               -> send Pong
             -- TODO:Handle them nicely
-            MessageInFailure _ -> shutdown
+            MessageInFailure _ -> return ()
```
ksaric: If the message that gets in fails, won't the sender of that message wait for it's response?
And if it is waiting for the response, won't that simply leave the initial process hanging?
Wouldn't it be nicer if the _sender_ got some response to know that things went bad?
`MessageOutFailure` with the wrong format or something similar?

src/Cardano/Shell/NodeIPC/Lib.hs
ksaric: At the end of the day, we will move these files into the executable anyway, so it's not really critical.

src/Cardano/Shell/NodeIPC/Lib.hs
HirotoShioi: I'm not understanding this comment well. This is when the sender is sending `MessageInFailure` message to the IPC so isn't it up to the sender what to do with their own process when they send this message to the node?

src/Cardano/Shell/NodeIPC/Example.hs
```diff
@@ -56,9 +56,9 @@ exampleWithFD = do
     (clientReadHandle, clientWriteHandle) <- getReadWriteHandles
     (serverReadHandle, serverWriteHandle) <- getReadWriteHandles
 
-    -- Start the server
     let nodePort = Port 8090
-    _ <- async $ startNodeJsIPC serverReadHandle clientWriteHandle nodePort
+    asyncServer <- async $ startIPC serverReadHandle clientWriteHandle nodePort
+    link asyncServer
```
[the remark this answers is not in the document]
HirotoShioi: If you do not use `link`, then a main thread will not cause an exception. To catch it, the main thread should use `race` or `concurrently`.

src/Cardano/Shell/NodeIPC/Message.hs
```diff
@@ -67,18 +69,19 @@ sendMessage (WriteHandle hndl) cmd = liftIO $ send $ encode cmd
...
+    encodedMessage <- liftIO $ readMessageFromHandle buildOS
     either
         (\_ -> throwM $ DecodeFail encodedMessage)
         return
         (eitherDecode encodedMessage)
   where
+    readMessageFromHandle :: OS -> IO BSL.ByteString
+    readMessageFromHandle = \case
+        Windows -> do
+            (_, _, blob) <- windowsReadMessage
```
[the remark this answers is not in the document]
HirotoShioi: Sorry, I do not understand this. How do you want this to be implemented?

src/Cardano/Shell/NodeIPC/Lib.hs
ksaric: Whops. Yes, you are correct, this is the message from the server.
But the message is saying "the last message I got was invalid"?
I guess this is good for now, since we really don't know what to do with that information.

src/Cardano/Shell/NodeIPC/Example.hs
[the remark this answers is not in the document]
ksaric: Indeed. I thought that the async exception will get propagated eventually unless you use `mask`. I learned something new.
```
testNoLink = do
    async $ do
        putTextLn "Hi there!"
        _ <- throwIO IPCException
        putTextLn "Some new line!"

testLink = do
    newAsync <- async $ do
        putTextLn "Hi there!"
        _ <- throwIO IPCException
        putTextLn "Some new line!"
    link newAsync
```
With:
```
Hi there!
```
And:
```
Hi there!
*** Exception: ExceptionInLinkedThread ThreadId 332 IOError has occured
```

src/Cardano/Shell/NodeIPC/Message.hs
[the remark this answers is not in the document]
ksaric: A bit more clearer:
```
readMessageFromHandle = \case
    Windows -> windowsReadMessage
    _       -> linuxReadMessage
```

PR #76, opened 2019-03-26
Put exlusive lock on application lock file

HirotoShioi:


ksaric:
I don't see a point in adding a library that does what the other 2 lines of code already did.

HirotoShioi:
So after talking with michael bishop, what I understand is that to implement this, I need to use `filelock` library which uses `FileLockEx` on Windows and that will safely put a exclusive lock on a lock file in Windows.

Issue #24, opened 2018-12-06
Add understandable documentation for non-developers


ksaric:
We need to write some understandable documentation for non-developers and the public.
The point of this is to have a more transparent and more efficient communication with non-developers and learn about the codebase in the process of doing that.

A good example of a document is https://github.com/input-output-hk/cardano-shell/blob/develop/ARCHITECTURE.md, but it starts to get a bit technical at some point (when we start speaking about layers).

PR #81, opened 2019-04-01
[GH-24] Add understandable documentation for non-developers

HirotoShioi:
Create an document explaining what the shell does for non-developers.

### DIFF
--- a/docs/Document.md
+++ b/docs/Document.md
@@ -1,0 +1,45 @@
+ # Overview
+
+`cardano-shell` is an thin layer which brings all the other modules working together and makes sure that they have the required resources (configuration, exception handling, monitoring, logging, ...).
+
+The shell is also responsible for:
+* Logging
+* Monitoring
+* Configuration
+* Exception handling
+* Launcher
+
+## Features
+
+Various **features** are required for the cardano node to operate.
+
+* Blockchain
+* Ledger
+* Logging
+* Monitoring
+* Network
+* Wallet backend
+
+The shell will act as an **glue** of these features. It provides required resources/configurations to each of these features as well as resolving its dependencies.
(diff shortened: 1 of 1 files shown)


docs/Document.md
```diff
@@ -0,0 +1,45 @@
...
+Various **features** are required for the cardano node to operate.
+
+* Blockchain
+* Ledger
+* Logging
+* Monitoring
+* Network
+* Wallet backend
+
+The shell will act as an **glue** of these features. It provides required resources/configurations to each of these features as well as resolving its dependencies.
```
denisshevchenko: > as an **glue**

as a **glue**, no?

HirotoShioi: Yes you're right. Will fix this.

docs/Document.md
```diff
@@ -0,0 +1,67 @@
+ # `cardano-shell` overview
+
+`cardano-shell` is an thin layer which brings all the other modules working
+ together and makes sure that they have the required resources 
+ (configuration, exception handling, monitoring, logging, ...).
+
+The shell is also responsible for:
```
ksaric: The shell is not responsible for logging and monitoring. That is brought in using the library, which is a part of shell (the part that can add features).

docs/Document.md
```diff
@@ -0,0 +1,67 @@
...
+Various **features** are required for the cardano node to operate.
+
+* Blockchain layer
+* Ledger layer
+* Logging layer
+* Monitoring layer
+* Network layer
+* Wallet backend layer
+
+The shell will act as a **glue** of these features. It provides required
```
ksaric: The idea is that it will be a glue _for some of them_.

PR #83, opened 2019-04-03
Update build to use nix-tools.

jbgi:
nix-shell is not functional yet.


jbgi:
Apparently there is some dependencies on `unix` that need to be guarded by an OS conditional in the cabal config, for cross-compilation to work.

CodiePP:
@Jimbo4350 also did some work on this: https://github.com/input-output-hk/cardano-shell/tree/jordan/nix-build-unix-windows

CodiePP:
bors try

ksaric:
bors try harder

CodiePP:
please, bors try

CodiePP:
`buildkite` now failing..

CodiePP:
rebased, regenerated nix files, update iohk-nix, and force pushed!
the cross-compilation is segfaulting (@angerman help!)
the command I used:
`nix build -f release.nix  nix-tools.libs.x86_64-pc-mingw32-cardano-shell.x86_64-linux`

angerman:
are you sure it's just cross that doesn't work @CodiePP?

Hydra paints a pretty bleak picture: https://hydra.iohk.io/jobset/Cardano/cardano-shell-pr-83#tabs-jobs

angerman:
Ahh ... lovely:
```
ghc: could not execute: /nix/store/47awn0vmnkffgzrhabhp8ylbxm3c1b59-ghc-8.6.3-with-contra-tracer/lib/ghc-8.6.3/bin/unlit
```
🤔

angerman:
I knew this sounded vaguely familiar: https://github.com/input-output-hk/haskell.nix/pull/78

Guess I'll have to look into this now.

ksaric:
The build doesn't work for me:
```
> stack build --nix -j12

trace: 
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
************************************ WARNING ***********************************
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

         You are using GHC 8.6.3.  This version is known to
         be busted on windows!  See GHC issue #16057.  Make
         sure you revert commit
           ghc:ed86e3b531322f74d2c2d00d7ff8662b08fabde6
         before using GHC 8.6.3 in any form on windows.

         --
         https://ghc.haskell.org/trac/ghc/ticket/16057

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
************************************ WARNING ***********************************
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

canonical-json-0.5.0.2: unregistering (local file changes: /home/ksaric/.stack/programs/x86_64-linux/ghc-8.6.3/lib/ghc-8.6.3/include/ghcversion.h /usr/inclu...)
canonical-json-0.5.0.2: build (lib)
profunctors-5.3: configure
canonical-json-0.5.0.2: copy/register
profunctors-5.3: build                                                     
semigroupoids-5.3.2: configure                                             
semigroupoids-5.3.2: build                                                 
threepenny-gui-0.8.3.0: configure                                          
threepenny-gui-0.8.3.0: build                                              
profunctors-5.3: copy/register                                             
semigroupoids-5.3.2: copy/register                 
Progress 4/21                     

--  While building package threepenny-gui-0.8.3.0 using:
      /home/ksaric/.stack/setup-exe-cache/x86_64-linux/Cabal-simple_mPHDZzAJ_2.4.0.1_ghc-8.6.3 --builddir=.stack-work/dist/x86_64-linux/Cabal-2.4.0.1 build --ghc-options " -ddump-hi -ddump-to-file -fdiagnostics-color=always"
    Process exited with code: ExitFailure 1
    Logs have been written to: /home/ksaric/projects/haskell/cardano-shell/.stack-work/logs/threepenny-gui-0.8.3.0.log

    Configuring threepenny-gui-0.8.3.0...
    Preprocessing library for threepenny-gui-0.8.3.0..
    Building library for threepenny-gui-0.8.3.0..
    [ 1 of 29] Compiling Foreign.JavaScript.Include ( src/Foreign/JavaScript/Include.hs, .stack-work/dist/x86_64-linux/Cabal-2.4.0.1/build/Foreign/JavaScript/Include.o )
    [ 2 of 29] Compiling Foreign.JavaScript.Resources ( src/Foreign/JavaScript/Resources.hs, .stack-work/dist/x86_64-linux/Cabal-2.4.0.1/build/Foreign/JavaScript/Resources.o )
    <command line>: can't load .so/.DLL for: libz.so (libz.so: cannot open shared object file: No such file or directory)
```

ruhatch:
@ksaric how are you building it? You need the `zlib` dependency from somewhere.

ksaric:
It's on the first line of the comment above, using stack.
Heh, from somewhere... I wish I knew where that is.

ksaric:
@CodiePP The "pure" Nix build is working using `nix build -f release.nix nix-tools.libs.x86_64-pc-mingw32-cardano-shell.x86_64-linux`.

ruhatch:
Ah, yes, good point - you can use something like the [stack-shell.nix](https://github.com/input-output-hk/cardano-ledger/blob/master/scripts/nix/stack-shell.nix)

ruhatch:
Then in `stack.yaml`

```
nix:
  shell-file: path/to/stack-shell.nix
```

ruhatch:
Oh, looks like you have that already, but it's different from ours and maybe doesn't have external dependencies

angerman:
@rvl may be able to assist with the stack/nix integration (or explanation why this doesn't work).

ksaric:
That would be great. 
I tried switching to the "other" `stack` file, clearing the `.stack-work` and tried modifying the configuration but the result is the same.

shell.nix
```diff
@@ -1,17 +1,9 @@
...
-    buildInputs = [ pkgs.haskellPackages.stack-hpc-coveralls pkgs.stack ];
-    shellHook = ''
-      echo '~~~ stack test'
-      stack test --coverage
-      echo '~~~ shc'
-      shc --repo-token=$COVERALLS_REPO_TOKEN cardano-shell cardano-shell-test
-      exit
-    '';
-  };
+  mainShell = localPkgs.nix-tools.libs.cardano-shell.env;
```
rvl: Change this to:
```suggestion
  mainShell = localPkgs.nix-tools.libs.cardano-shell;
```
It should really be `pkgSet.config.hsPkgs.cardano-shell.components.all` though.

stack-shell.nix
```diff
@@ -1,21 +1,4 @@
...
-
-in
-  with pkgs;
-
-  haskell.lib.buildStackProject {
-    name = "cardano-shell-env";
-    buildInputs = [ zlib openssl git ];
-    ghc = haskell.packages.ghc863.ghc;
-  }
+(import ./default.nix { inherit system config; }).stack-env
```
rvl: * Why not keep the `haskell.lib.buildStackProject` in here?
* Import `default.nix` and use it to get the pinned `pkgs`.
* Remove the `system` and `config` args from this file because they aren't relevant for nix-shell (this was originally my mistake).
* You should be able to get the correct version of `ghc` from the haskell.nix package set.

nix/iohk-nix-src.json
```diff
@@ -0,0 +1,7 @@
+{
+  "url": "https://github.com/input-output-hk/iohk-nix",
+  "rev": "0a34a05d611a7141d2b6d73446be004fa06f0f5c",
```
rvl: Update `haskell.nix` to the latest revision on `master` and nix-shell development environments should work. @jbgi

.stack-to-nix.cache
```diff
@@ -0,0 +1,12 @@
+https://github.com/input-output-hk/cardano-prelude 2256fd727c5f92e6218afdcf8cddf6e01c4a9dcd . 0fdwlhkpc7inkqflcdzinx9qr5g3i34clzhl6iiagj851c3jcgsn cardano-prelude cardano-prelude.nix
```
rvl: I would remove this file from git because it often causes problems when regenerating with stack-to-nix.

angerman:
> This PR it is now very close to a state when we can actually merge and close it.
> 
> working: cross-compilation with `nix-tools`
> not working: `stack build --nix` (which just requires a build dependency of `libz` somewhere)
> 
> who could do this very little last change?

What is the actually expected behavior of `stack build --nix`?

ksaric:
> What is the actually expected behavior of stack build --nix?

That it builds? Loosely `stack build --nix` == "build with stack using nix dependencies"?
From the https://docs.haskellstack.org/en/stable/nix_integration/ - "Nix is used by Stack to provide the non-Haskell dependencies needed by these Haskell packages."
The rest should be from Stackage.
I tried playing with this and I can no longer track what we are doing with Stack, Nix, Cabal, nix-tools, custom nix derivations and the rest.

angerman:
> > What is the actually expected behavior of stack build --nix?
> 
> That it builds? Loosely `stack build --nix` == "build with stack using nix dependencies"?
> From the https://docs.haskellstack.org/en/stable/nix_integration/ - "Nix is used by Stack to provide the non-Haskell dependencies needed by these Haskell packages."
> The rest should be from Stackage.
> I tried playing with this and I can no longer track what we are doing with Stack, Nix, Cabal, nix-tools, custom nix derivations and the rest.

So if it's just about stack using system dependencies, wouldn't 
```yaml
nix:
  enable: true
  packages: [zlib]
```
be enought in the `stack.yaml`?6

ksaric:
It will report:

> You cannot have packages and a shell-file filled at the same time in your nix-shell configuration.

That is a Nix dependency and should be pulled from it, but for some unknown reason, it's not found.

shell.nix
ksaric: Neither of these does the trick, though.

angerman:
@ksaric I meant to suggest to replace:
```yaml
nix:
  enable: true
  pure: true
  shell-file: shell.nix
```
with
```
nix:
  enable: true
  packages: [zlib]
```
assuming `zlib` is the only system dep needed. That would be as declarative as possible and let `stack` perform the necessary logic for stack/nix integration.

Coming back to your earlier question:
> I can no longer track what we are doing with Stack, Nix, Cabal, nix-tools, custom nix derivations and the rest.

We use `nix` for CI, and the CI system is `hydra`, which can evaluate nix-expressions. I'm not going to go into too much detail here. But the bottom line is that there are two appraoches of building haskell packages, one is using the build infrastructure that comes with `nixpkgs`, the other is what we built and call `haskell.nix` (or somtimes also referred to as `nix-tools`).  The approach in `nixpkgs` will build haskell projects as a set of packages. Whereas `haskell.nix` will build the project as a set of package components. Thus the granularity of `haskell.nix` is improved over the one in `nixpkgs`. This comes with lots of additional benefits. The downside is that `haskell.nix` is only getting slowly more traction in the community and also does not come with all the features that have already been introduced into the `nixpkgs` infrastructure. So why a new one anyway?  I can't tell you how much I would have loved *not* to write `haskell.nix`.  However the existing one in `nixpkgs` is very ill suited for cross compilation, proper stackage handling (iohk even wrote a `stack2nix` tool to try to make the existing infrastructure deal with stack projects), or even cabal project handling.

stack-shell.nix
jbgi: so with `buildStackProject` I got this error when running `satck build`:
```
/home/jbgi/.stack/setup-exe-cache/x86_64-linux-nix/Cabal-simple_mPHDZzAJ_2.4.0.1_ghc-8.6.3: startProcess: runInteractiveProcess: exec: does not exist (No such file or directory)
```
any idea?

jbgi:
@CodiePP `stack build` is fixed now. But is not completely satisfactory because the nix-shell still depends on `cardano-shell` lib components instead of only its dependencies (this will be fixed via https://github.com/input-output-hk/haskell.nix/pull/121)

ksaric:
@jbgi When running stack (`stack build --fast --nix`) I get:
```
Cloning cbe7ab32354f3838dc8c95c64109904c8f503347 from https://github.com/input-output-hk/iohk-monitoring-framework
Executable named git not found on path
```

jbgi:
@ksaric thanks, I missed that due to not trying from a fresh checkout.
While waiting for https://github.com/input-output-hk/haskell.nix/pull/121 I have disabled purity for the nix-shell used by stack so that stack can pick-up git for the user environment.

stack.yaml
```diff
@@ -96,4 +100,6 @@ extra-deps:
 # compiler-check: newer-minor
 
 nix:
+  enable: true
+  pure: false
```
ksaric: It should be pure if all the dependencies are provided from the Nix environment, right?

jbgi: right, currently git is not provided. It cannot be added easily until we get input-output-hk/haskell.nix#121.

Issue #78, opened 2019-03-28
Create the initial launcher version


ksaric:
After #70, we can implement the basic launcher which can start up Daedalus.

PR #84, opened 2019-04-05
[GH-78] Create the initial launcher version.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/78

It turns out that this simple launcher is good enough for now, we have yet to define the configuration.


HirotoShioi:
I got two exceptions:
1. When running `main`, it throws exception saying directory `wallet` does not exist.
`*** Exception: ./wallet: createProcess: runInteractiveProcess: exec: does not exist (No such file or directory)`
2. When I create directory `wallet` and run again, I still get exception saying that it does not have permission.
`*** Exception: ./wallet: createProcess: runInteractiveProcess: exec: permission denied (Permission denied)`

Issue #85, opened 2019-04-05
Generalize node IPC so it can be extended


ksaric:
In a chat with @KtorZ, he proposed that we make the IPC protocol open to an extension so that they can add additional functionality/messages if required.
Having thought about that, I think we can do that, given that we agreed that the communication has to be one-to-one (one message in - one message out).

PR #89, opened 2019-04-10
[GH-78] Generalize node IPC so it can be extended.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/85

I can't say I really managed to create what I planned to do, but it's still not bad - the protocol handling function is extracted and I have fixed the IPC behavior when running for a longer time (more than a single message).

### DIFF
--- a/src/Cardano/Shell/NodeIPC/Lib.hs
+++ b/src/Cardano/Shell/NodeIPC/Lib.hs
@@ -3,16 +3,16 @@
 <https://github.com/input-output-hk/cardano-shell/blob/develop/specs/CardanoShellSpec.pdf>
 -}
 
-{-# LANGUAGE DeriveGeneric       #-}
-{-# LANGUAGE LambdaCase          #-}
-{-# LANGUAGE OverloadedStrings   #-}
-{-# LANGUAGE RankNTypes          #-}
-{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE DeriveGeneric          #-}
+{-# LANGUAGE LambdaCase             #-}
+{-# LANGUAGE RankNTypes             #-}
+{-# LANGUAGE ScopedTypeVariables    #-}
 
 module Cardano.Shell.NodeIPC.Lib
     ( startNodeJsIPC
     , startIPC
     , Port (..)
+    , ProtocolDuration (..)
     -- * Testing
     , getIPCHandle
     , MsgIn(..)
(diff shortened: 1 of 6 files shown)


src/Cardano/Shell/NodeIPC/Lib.hs
```diff
@@ -52,6 +52,14 @@ import           Cardano.Shell.NodeIPC.Message (MessageException,
...
+-- | The way the IPC protocol works.
+data ProtocolDuration
+    = SingleMessage
+    -- ^ Responds to a single message and exits
+    | MultiMessage
+    -- ^ Runs forever responding to messages
+    deriving (Eq, Show)
+
 -- | Message expecting from Daedalus
 data MsgIn
```
ksaric: Messages that have to be defined.

src/Cardano/Shell/NodeIPC/Lib.hs
```diff
@@ -162,44 +176,58 @@ getIPCHandle = do
...
+startNodeJsIPC :: forall m. (MonadIO m) => ProtocolDuration -> Port -> m ()
+startNodeJsIPC protocolDuration port = do
+    handle          <- liftIO $ getIPCHandle
+    let readHandle  = ReadHandle handle
     let writeHandle = WriteHandle handle
-    liftIO $ void $ ipcListener readHandle writeHandle port
+    liftIO $ void $ ipcListener protocolDuration readHandle writeHandle port
+
+-- | Function for handling the protocol
+handleIPCProtocol :: forall m. (MonadIO m) => Port -> MsgIn -> m MsgOut
```
ksaric: The protocol handler function, which can be extended.
The `Port` is there because it's the simplest way to include it.

src/Cardano/Shell/NodeIPC/Lib.hs
```diff
@@ -162,44 +176,58 @@ getIPCHandle = do
...
-            MessageInFailure _ -> return ()
+        send Started -- Send the message first time the IPC is up!
+
+        let frequencyFunction = case protocolDuration of
+                                    SingleMessage -> void
+                                    MultiMessage  -> forever
+
+        -- Fetch message and respond to it
+        frequencyFunction $ do
+            msgIn               <- readMessage readHandle        -- Read message
```
ksaric: The main reason why I can't generalize the function, we don't know what the type is.

src/Cardano/Shell/NodeIPC/Lib.hs
HirotoShioi: Does extended mean you can have the shell to perform any arbitrary `IO` action when handling the message? If so, this LGTM.
At glance I didn't understand why this function has MonadIO constraint

ksaric: > At glance I didn't understand why this function has MonadIO constraint

It's there to enable adding any kind of message to the protocol.
So if somebody first needs to fetch something or do something with `IO` before returning a response, they have nothing to fear (except the amount of freedom).

PR #90, opened 2019-04-24
[GH-99] T5.1.1 added configuration to logging feature

CodiePP:
added access to logging configuration in `LoggingFeature`

needed by https://github.com/input-output-hk/iohk-monitoring-framework/issues/228

issue #99

basis for `cardano-ledger` PR: https://github.com/input-output-hk/cardano-ledger/pull/487

### DIFF
--- a/app/Cardano/Shell/Features/Logging.hs
+++ b/app/Cardano/Shell/Features/Logging.hs
@@ -9,6 +9,7 @@ module Cardano.Shell.Features.Logging
     -- re-exports
     , Trace
     , Configuration
+    , LoggerName
     , Severity (..)
     ) where
 
--- a/app/Cardano/Shell/Features/Logging.hs
+++ b/app/Cardano/Shell/Features/Logging.hs
@@ -48,13 +48,14 @@ data LoggingParameters = LoggingParameters
 -- the functions effects and constraining the user (programmer) of those function to use specific effects in them.
 -- https://github.com/input-output-hk/cardano-sl/blob/develop/util/src/Pos/Util/Log/LogSafe.hs
 data LoggingLayer = LoggingLayer
-    { llBasicTrace :: forall m . MonadIO m => Trace m Text
-    , llLogDebug   :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
-    , llLogInfo    :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
-    , llLogNotice  :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
-    , llLogWarning :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
-    , llLogError   :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
-    , llAppendName :: forall m a. (MonadIO m, Show a) => LoggerName -> Trace m a -> m (Trace m a)
+    { llBasicTrace    :: forall m . MonadIO m => Trace m Text
+    , llLogDebug      :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
+    , llLogInfo       :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
+    , llLogNotice     :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
+    , llLogWarning    :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
+    , llLogError      :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
+    , llAppendName    :: forall m a. (MonadIO m, Show a) => LoggerName -> Trace m a -> m (Trace m a)
+    , llConfiguration :: Configuration
     }
 
 --------------------------------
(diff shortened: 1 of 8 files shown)


app/Cardano/Shell/Features/Logging.hs
```diff
@@ -94,13 +95,14 @@ loggingCardanoFeatureInit loggingConfig = do
...
-                    , llLogError    = Trace.logError
-                    , llAppendName  = Trace.appendName
+                    { llBasicTrace    = Trace.natTrace liftIO baseTrace
+                    , llLogDebug      = Trace.logDebug
+                    , llLogInfo       = Trace.logInfo
+                    , llLogNotice     = Trace.logNotice
+                    , llLogWarning    = Trace.logWarning
+                    , llLogError      = Trace.logError
+                    , llAppendName    = Trace.appendName
+                    , llConfiguration = lpConfiguration loggingConfig
```
ksaric: This looks kind of weird. 
Isn't it possible to add the `bracket` function here and export it?

CodiePP: yes, I am experimenting with exporting everything that is needed in the target app and get rid of the direct import of the logging library. Types and functions should be (re-)exported through the feature in `node-shell`.

app/Cardano/Shell/Features/Logging.hs
```diff
@@ -48,13 +56,16 @@ data LoggingParameters = LoggingParameters
...
-    , llLogError   :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
-    , llAppendName :: forall m a. (MonadIO m, Show a) => LoggerName -> Trace m a -> m (Trace m a)
+    { llBasicTrace      :: forall m . MonadIO m => Trace m Text
+    , llLogDebug        :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
+    , llLogInfo         :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
+    , llLogNotice       :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
+    , llLogWarning      :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
+    , llLogError        :: forall m a. (MonadIO m, Show a) => Trace m a  -> a -> m ()
+    , llAppendName      :: forall m a. (MonadIO m, Show a) => LoggerName -> Trace m a -> m (Trace m a)
+    , llConfiguration   :: Configuration
```
ksaric: Like I said, I don't think we need a `Configuration`.

app/Cardano/Shell/Features/Logging.hs
```diff
@@ -48,13 +56,16 @@ data LoggingParameters = LoggingParameters
...
+    , llBracketMonadX   :: forall m a t. (MonadIO m, Show a) => Configuration -> Trace m a -> Severity -> Text -> m t -> m t
```
ksaric: Formatting is messed up.

app/Cardano/Shell/Features/Logging.hs
```diff
@@ -48,13 +56,16 @@ data LoggingParameters = LoggingParameters
...
+    , llBracketStmIO    :: forall a t. (Show a) => Configuration -> Trace IO a -> Severity -> Text -> STM.STM t -> IO t
```
ksaric: Formatting.

cardano-shell.cabal
```diff
@@ -45,30 +45,28 @@ library
     , specs
   build-depends:
       aeson
-    , base >=4.7 && <5
+    , base >=4.11 && <5
```
ksaric: Why is this changed?

cardano-shell.cabal
```diff
@@ -45,30 +45,28 @@ library
...
     , binary
     , bytestring
-    , cardano-prelude
     , Cabal
+    , cardano-prelude
```
ksaric: I don't like the fact that people are reordering dependencies and removing comments from them.

cardano-shell.cabal
```diff
@@ -45,30 +45,28 @@ library
...
-    , dhall
-    , ekg-core
     , process
     , QuickCheck
+    , safe-exceptions
+    , stm
     , text
     , transformers
-    , unix
+  if os(windows)
```
ksaric: Good point, the path we were going with was "let's have things working on *nix and we'll deal with Windows later".

src/Cardano/Shell/Types.hs
```diff
@@ -17,7 +17,7 @@ import           Control.Concurrent.Classy (MonadConc)
 
 import           Cardano.Shell.Constants.Types (CardanoConfiguration (..))
 
-import qualified System.Metrics as Ekg
+-- import qualified System.Metrics as Ekg
```
ksaric: Do we need these comments? EKG is not really an issue, the stack traces just indicate something went wrong, not that it's crashing the whole application.

erikd: Commented out code should never be merged to master.

app/Cardano/Shell/Features/Logging.hs
erikd: The code in `cardano-ledger` currently uses the `llConfiguration` field.

cardano-shell.cabal
erikd: Most of the `cardano-*` projects only build with ghc-8.4 and later. Ghc-8.4 ships with `base-4.11.*`

CodiePP: I tried to get rid of the contravariant package. It is included in base-4.12.
But, couldn't raise the version too high. Can be reverted, unless prelude is updated.

cardano-shell.cabal
CodiePP: this is required by the iohk-nix cross-compilation setup in cardano-ledger.

PR #145, opened 2019-05-09
[GH-144] Rethrow cought exceptions to check if the application is throwing them.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/144

This should be released to test with `validate-mainnet`.
Also, other exceptions should be noticed.

After confirming the exception is not swallowed, we can continue.


erikd:
Seems sensible, especially to do this now rather than later.

Most exceptions should be caught as close as possible to their source and either handled in an appropriate manner or returned back up the stack as `Left error`. Having a global top level exception handler than just retries things can hide nasty errors.

Issue #86, opened 2019-04-08
Create one-pager for cardano-shell


HirotoShioi:
We want one pager like this so that other teams can understand what the shell is.
https://github.com/input-output-hk/iohk-monitoring-framework/blob/master/docs/Logging_one-pager.pdf

PR #146, opened 2019-05-10
[GH-86] Add one-pager pdf

HirotoShioi:
This PR introduces one-pager pdf file for cardano-shell.
https://github.com/input-output-hk/cardano-shell/issues/86


erikd:
PDF is a binary format that cannot be meaningfully diffed or otherwise handled the way we handle source code. That means all we can ever provide is snapshots of this file, rather than any meaninigful information as to how it has changed over time.

I would therefore suggest that we do not commit the PDF file itself to git, but rather the hopefully human readable source file, eg markdown or LaTeX.

HirotoShioi:
@erikd I agree with you. I've reported this to the @CodiePP and he'll look into it.

PR #148, opened 2019-05-13
Update document

HirotoShioi:
Updated NodeIPC documentation to include description to include about extended message handling.


ksaric:
@HirotoShioi When you see a failure, you can take a look at the "Details" and try to figure out why is it failing.

specs/CardanoShellSpec.lhs
```diff
@@ -207,6 +207,9 @@ When the \textit{Daedalus} calls and starts the \textit{Node}, it also opens up
 First, the \textit{Node} sends the message \textbf{Started} back to the \textit{Daedalus} to inform him that the communication can begin.
 After that, \textit{Daedalus} sends the message \textbf{QueryPort} to the \textit{Node}, and the \textit{Node} responds with the free port it found using \textbf{ReplyPort PORTNUM} that is going to be used for starting the HTTP "server" serving the \textit{JSON API} which they can then use to communicate further.\\
 
+Not only does the \textit{Node} responds to the message, but it can also perform some IO actions depending on the message sent from the client.
+For instance, you can have the node to start an new process when the client sends \textbf{Started} or have it so that node will kill its thread/process when it responds with \textbf{ReplyPort PORTNUM}.\\
```
ksaric: `start an new process` - `start a new process`.
"you can have the node to start an new process when the client sends \textbf{Started} or have it so that node will kill its thread/process when it " - what does this mean? Maybe an example where the response is an action called "FetchInfo SystemInfo" where the client needs to fetch the client info and return it to the server. That would require `IO`.

ksaric:
You can rebase on top of https://github.com/input-output-hk/cardano-shell/pull/171

specs/shell.nix
```diff
@@ -1,6 +1,6 @@
 let
   # adding the haskell environment for running lhs2tex
-  pkgs = import (import ./../nixpkgs.nix) { config = import ./../config.nix; };
+  pkgs = import (import ./../nixpkgs.nix) { config = import ./../default.nix; };
```
ksaric: I'm not sure if we even need `nixpkgs.nix` but this is not required to be the bleeding edge anyway, so it's good.

Issue #149, opened 2019-05-13
When running the IPC examples `System.Posix.Process` does not exist on Windows


ksaric:
The file `src/Cardano/Shell/NodeIPC/Example.hs` uses `System.Posix.Process.forkProcess`
traditional POSIX forking which was used temporarily for the example.

It should be switched to use the traditional `process` and `System.Process` from Haskell which should be cross-compile compatible.

CodiePP:
maybe start from branch `nix-tools-update`

check that cross-compilation works:
`nix build -f release.nix nix-tools.libs.x86_64-pc-mingw32-cardano-shell.x86_64-linux`

ksaric:
@HirotoShioi Try to provide the example for the Haskell-Haskell process communication which you said doesn't work, otherwise, we can't know whether it works or not. We never saw that example.

HirotoShioi:
@CodiePP I tried everything and still couldn't compile it.
As a last resort, I've used docker with nix installed (https://hub.docker.com/r/nixos/nix/) and tried to compile it and it failed with this error:

Steps:
1. Run the image `docker run -it nixos/nix`
2. Install git `apk add git`
3. Clone this repo, `git clone https://github.com/input-output-hk/cardano-shell.git`
4. Checkout to the branch `git checkout nix-tools-update`
5. Run the script: `nix build -f release.nix nix-tools.libs.cardano-shell.x86_64-linux`

```
[...]

PR #152, opened 2019-05-15
Remove the use of System.Posix.Process, use System.Process instead

HirotoShioi:
In `Example.hs` Use `System.Process` instead of `System.Posix.Process` to enable cross-compilation.
https://github.com/input-output-hk/cardano-shell/issues/149


HirotoShioi:
I've pointed wrong branch.. Recreating the PR

PR #153, opened 2019-05-15
Use process library, remove unix from dependency

HirotoShioi:
In `Example.hs`, Use `System.Process` instead of `System.Posix.Process` to enable cross-compilation.
https://github.com/input-output-hk/cardano-shell/issues/149

How to run:
1. Have everything built beforehand using `stack`.
2. Run repl, enter command `exampleWithProcess Ping`


HirotoShioi:
Runnning ` ./nix/regenerate.sh` to apply the patch in the buildkite artifact
https://buildkite.com/input-output-hk/cardano-shell/builds/531#07812c03-ce8e-4e9c-ba5a-91a8f7d26f5c

erikd:
@HirotoShioi I would recommend against running  `./nix/regenerate.sh` yourself. Instead, git the patch file that CI generates and apply that. The patch file is under the "artifacts" tab of the "stack2nix" job.

HirotoShioi:
@erikd I would do that! Thank you.

app/NodeIPC/Main.hs
```diff
@@ -1,12 +1,57 @@
...
+port = Port 8090
 
 main :: IO ()
 main = do
-    let port = Port 8090
-    startNodeJsIPC SingleMessage port
+    (cmd:_) <- getArgs
+    case cmd of
+        "js"      -> startNodeJsIPC SingleMessage port
+        "haskell" -> bracket acquire restore action
```
ksaric: Good.

app/NodeIPC/Main.hs
```diff
@@ -1,12 +1,57 @@
...
+    restore (clientWHandle, serverRHandle, serverWHandle) = do
+        hClose clientWHandle
+        hClose serverRHandle
+        hClose serverWHandle
+
+    action :: (Handle, Handle, Handle) -> IO ()
+    action (clientWHandle, serverRHandle, serverWHandle) = do
+        let clientWriteHandle = WriteHandle clientWHandle
+        let serverReadHandle  = ReadHandle serverRHandle
+        sendMessage (WriteHandle serverWHandle) Ping
```
ksaric: Should be the other way around, right?

src/Cardano/Shell/NodeIPC/Example.hs
```diff
@@ -59,42 +59,52 @@ exampleWithFD = do
 
     (clientReadHandle, clientWriteHandle) <- getReadWriteHandles
 
-    (_, responses) <-
-        ipcServer clientWriteHandle Ping
+    (_, responses) <- do
+        (serverReadHandle, serverWriteHandle) <- getReadWriteHandles
+        -- Send message to server
+        sendMessage serverWriteHandle Ping
```
ksaric: You can parametrize the message so the user can choose whether to send `Ping`, `QueryPort` or something else we add.

src/Cardano/Shell/NodeIPC/Example.hs
```diff
@@ -59,42 +59,52 @@ exampleWithFD = do
...
+        unsetEnv "FD_WRITE_HANDLE"
+
+    action :: (ReadHandle, Handle) -> IO (MsgOut, MsgOut)
+    action (readHandle, _) = do
+        withCreateProcess (proc "stack" ["exec", "node-ipc", "haskell"]) $
+        -- Tried to use stdin, but it is very incosistent.
+        -- the executable throws exception:
+        -- node-ipc: Unable to read with given handle: {handle: <stdout>}
+        -- Even though I'm providing stdin
+            \_ _ _ _ -> receieveMessages readHandle
```
ksaric: If you want to send the message once, you can use `void`, otherwise, you can use `forever`.

app/NodeIPC/Main.hs
```diff
@@ -1,12 +1,57 @@
...
+        startIPC SingleMessage serverReadHandle clientWriteHandle port
+
+    getHandleFromEnv :: String -> IO Handle
+    getHandleFromEnv envName = do
+        mFdstring <- lookupEnv envName
+        case mFdstring of
+            Nothing -> error $ "Unable to find fd: " <> envName
+            Just fdstring -> case readEither fdstring of
+                Left err -> error $ "Could not parse file descriptor: " <> toS err
```
erikd: I am not a big fan of error handling via just throwing an `error`. It works acceptable well for small projects and demos, but scales very poorly to larger projects where you can suddenly have 100s of calls to `error`.

I think it would be much better to have the type of `getHandleFromEnv` to be:
```
    getHandleFromEnv :: String -> IO (Either NodeIPCError Handle)
```
and then have `NodeIPCError` be a sum type with a proper render function (`NodeIPCError -> Text`) that can be used to print the error.

ksaric: "The illusion of certainty" - using either the `error` or the `Either` here will result in the application "crashing", right?
In either case, it's a non-recoverable exception, right - does `Either` bring something here other than the delay of the exception - the handler of the `Either` will have to do something like `error` or `panic`?
It's nice for testing, having "pure" exceptions, reasoning about the function via its type and it might be easier to reason about in some cases, but sometimes the result is the same - when you have a `Left`, you shutdown with an error message.
While I agree in principle, I always wonder what/where is the difference, since the end result is the same in cases like these? It does make things complicated when you have it passing that signature around and at some point, you do the same thing you could have done in the first place.
:hear_no_evil:

erikd: @ksaric I would agree for how this is structured at the moment, as a standalone application.

However, my experience suggests that anything other than a demo should be a library and a minimal wrapper to turn it into an application. Ideally, this thin wrapper should contain nothing more than command line parsing and calls into the library.

I hope its rather obvious that library code should never call `error` (or more controversially, throw an exception), but rather pass errors back to the application, preferably as the `Left` of an `Either`. 
The library of course should not expose functions that use `ExceptT`, but certainly can use `ExceptT` internally.

The reasons to make libraries:
* Code designed as a library is usually better structured than application code and will usually have a real API.
* Code build as a library for one client program can be easily used by another client, probably without rework.
* Code built as a library is easier to test than code built as an application.

The reasons to use `ExceptT` rather than exceptions:
* With `ExceptT` the errors a function can have are explicitly listed in the type signature.
* Errors types (used with `Either` and `ExceptT`) are closed sets whereas the exception system was designed to be (and is) extensible.
* Errors types (used with `Either` and `ExceptT`) compose (data types within data types within data types ....) and the outer error data type is still a closed set. The exceptions system allows composition via its extensibility, but the open extensibility means there is no way for the compiler to ensure that all exceptions are handled.

ksaric: That is a very good point indeed. 
Visibility/Explicitness is important, especially when you want to use something that you didn't write yourself and don't know the execution paths.

HirotoShioi:
Yep, I agree, I should definitely polish the exception handling. I was rushing to make the demo work.
And I'm a bit concerned about the CI. I don't think the CI is running the test. I'm going to take a look at it tomorrow.

HirotoShioi:
So I finally have the CI running tests, and it seems like there's no `stack` available on the CI.

```
  test/NodeIPCSpec.hs:135:13: 
  1) NodeIPC.startIPC.Examples should return Started, Pong with createProcess
       uncaught exception: IOException of type NoSuchThing
       stack: createProcess: runInteractiveProcess: exec: does not exist (No such file or directory)
       (after 1 test)
```

I'm currently compiling with nix to see if there's any workaround.

app/NodeIPCClient/Main.hs
```diff
@@ -0,0 +1,54 @@
...
+    (cmd:_) <- getArgs
+    case cmd of
+        "js"      -> startNodeJsIPC SingleMessage port
+        "haskell" -> bracket acquire restore action
+        _         -> return ()
+  where
+    acquire :: IO Handle
+    acquire = do
+        -- Lookup the Handle that the server has set
+        serverWHandle <- getHandleFromEnv "FD_WRITE_HANDLE"
```
ksaric: I have to admit I'm confused by this one.
I understand why you put it in, but I'm not sure we need it.

cardano-shell.cabal
```diff
@@ -36,9 +36,12 @@ library
     , Cardano.Shell.Presets
     , Cardano.Shell.Features.Logging
   other-modules:
-      Cardano.Shell.NodeIPC.Example
-    , Cardano.Shell.NodeIPC.Lib
+      Cardano.Shell.NodeIPC.Lib
     , Cardano.Shell.NodeIPC.Message
+  if !os(windows)
```
ksaric: I don't think we need this.

shell.nix
```diff
@@ -1,3 +1,7 @@
+{ system ? builtins.currentSystem
```
ksaric: Is this rebased? Where did you pick this up?

src/Cardano/Shell/NodeIPC/Example.hs
```diff
@@ -15,25 +15,30 @@
 -- This allows the two proccesses to send the message to the other while
 -- reading the response that other had sent.
 -}
+
 module Cardano.Shell.NodeIPC.Example
```
ksaric: Should be called `ServerExample` or something like that, right?

HirotoShioi:
Created new PR which points to the `develop` branch.
https://github.com/input-output-hk/cardano-shell/pull/173

PR #168, opened 2019-05-27
Fix various `CardanoConfiguration` types

Jimbo4350:
- When parsing `CardanoConfiguration` to see if we require a `NetworkMagic` value, we look for either `"RequiresMagic"` (for testnet) or `"RequiresNoMagic"` (for mainnet or staging).

- Change other types in `CardanoConfiguration` to match their corresponding types in `cardano-ledger`.


ksaric:
You can rebase on top of https://github.com/input-output-hk/cardano-shell/pull/171

PR #174, opened 2019-05-31
Use new shellFor function to provide build env.

jbgi:
Only build dependencies, allow pure stack build.
(followup on https://github.com/input-output-hk/cardano-shell/pull/83#issuecomment-494381029)

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -60,12 +60,12 @@ in defaultNix // {
     # env will provide the dependencies of cardano-shell
     packages = ps: with ps; [ cardano-shell ];
     # This adds git to the shell, which is used by stack.
-    buildInputs = with pkgs; [ git ];
+    buildInputs = with pkgs; [ git stack commonLib.stack-hpc-coveralls ];
   };
 
   runCoveralls = pkgs.stdenv.mkDerivation {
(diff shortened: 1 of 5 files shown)


default.nix
```diff
@@ -41,14 +41,27 @@ in
...
+  defaultNix = commonLib.nix-tools.default-nix ./nix/pkgs.nix { 
+    inherit system crossSystem config pkgs;
+  };
+in defaultNix // {
+  # ... and add additional packages we want to build on CI:
+
+  env = defaultNix.nix-tools.shellFor {
+    inherit withHoogle;
+    # env will provide the dependencies of cardano-shell
+    packages = ps: with ps; [ cardano-shell ];
```
ksaric: Do we have the stack coveralls package dependendency somewhere in the default package set? If not, it would be useful, since we could all have the same version and run the same thing.

jbgi: We do (defined in `iohk-nix`). I added it to the default nix-shell.

jbgi:
@ksaric I added `stack-shell.nix`to avoid the unneeded weight of launching hoogle when building with stack. Hoogle is enable by default in the normal `nix-shell` case.

Issue #169, opened 2019-05-27
Storage and validation of AVVM UTxO


Jimbo4350:
- Determine the best way to store and validate the AVVM UTxO 
- How to incorporate it in the `cardano-shell` config (`CardanoConfiguration`)

ksaric:
Related to https://github.com/input-output-hk/cardano-shell/issues/151.

ksaric:
So the conclusion is that the parsing and validation of the whole genesis file are going to be handled in the `cardano-ledger`, while the `cardano-shell` can provide the path to the location and source of the Genesis files.
Also, something to keep in mind is that the old AVVM configuration section found here - https://github.com/input-output-hk/cardano-sl/blob/a566b5198d27ddaddef095d217b7578f3310a417/lib/configuration.yaml#L417 is already included in `mainnet-genesis.json` so it's not required to be separate.

ksaric:
@KtorZ was kind enough to provide the hashes of the files: 
- mainnet: 89d9b5a5b8ddc8d7e5a6795e9774d97faf1efea59b2caf7eaf9f8c5b32059df4
- staging: b365f1be6863b453f12b93e1810909b10c79a95ee44bf53414888513fe172c90
- testnet: 7f141ea26e189c9cb09e2473f6499561011d5d3c90dd642fde859ce02282a3ae

And previous blocks:
- mainnet: 5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb
- staging: c6a004d3d178f600cd8caa10abbebe1549bef878f0665aea2903472d5abf7323
- testnet: b7f76950bc4866423538ab7764fc1c7020b24a5f717a5bee3109ff2796567214

The hashes are blake2b 256 hash of the JSON-serialized genesis config file (in canonical form).

And @robcohen provided the source of the actual hash, documented - https://github.com/input-output-hk/cardano-sl/blob/4f5138d4390e5c55e2b6db3d4d6f113d076513a1/docs/configuration.md#how-to-specify-genesis

ksaric:
Can be checked out by https://github.com/input-output-hk/cardano-sl/blob/798ddf09a5a2d14cabf4f8f417a5c81bffc9bff0/scripts/js/genesis-hash.js

PR #175, opened 2019-06-03
[GH-169] Storage and validation of AVVM UTxO, revised.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/169

Switched to just using the names of the genesis files and the required hashes (for the validation), the AVVM UTxO is already included there.


src/Cardano/Shell/Constants/Types.hs
```diff
@@ -65,11 +65,20 @@ data Core = Core
...
+-- __genesis config ---> genesis file__
+-- And separately:
+-- __genesis file ---> runtime config ---> running node__
+-- __static config ---> ...__
+--
+data Genesis = Genesis
+    { geSrc             :: !FilePath
+    , geGenesisHash     :: !Text
+    , gePrevBlockHash   :: !Text
+    } deriving (Eq, Show)
```
Jimbo4350: So my understanding is that we also need the `Spec` if we want to launch the node with a custom configuration. Do we really want to remove this? What is the alternative?

ksaric: That is something I was bothered with as well.
For example, we don't have the `k` parameter unless we have a "custom" configuration. And that "custom" configuration contains some parameters I'm not sure are required.
What I can say for now is that it seems we don't need them right away, and thus I simply removed them. I want to see what ends up missing rather than have some additional params which serve no use at all.
The types remain and if you do end up requiring specific parameters, I will return them easily.

PR #178, opened 2019-06-05
Remove unix library, replace with process

HirotoShioi:
Same PR as #153 but points to develop branch instead of nix-tools-update

How to run:

Have everything built beforehand using stack.
Run repl, enter command `exampleWithProcess Ping`


src/Cardano/Shell/NodeIPC/ServerExample.hs
```diff
@@ -0,0 +1,128 @@
...
+    restore ((ReadHandle rHandle), wHandle) = do
+        hClose rHandle
+        hClose wHandle
+        unsetEnv "NODE_CHANNEL_FD"
+
+    action :: MsgIn -> (ReadHandle, Handle) -> IO (MsgOut, MsgOut)
+    action msgin (readHandle, _) = do
+        withCreateProcess (proc "stack" ["exec", "node-ipc", "haskell"])
+            { std_in = CreatePipe } $
+                \(Just stdIn) _ _ _ -> do
```
ksaric: I don't like this since it's not really in sync with the JS side, but fine for now I guess.

Issue #160, opened 2019-05-17
Add shutdown message for node-IPC


ksaric:
Currently, Daedalus is using the http protocol to shut down the node.
We can include the shutdown in the IPC so they can use the IPC signal.

PR #185, opened 2019-06-07
[GH-160] Add shutdown message for node-IPC.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/160

### DIFF
--- a/test/NodeIPCSMSpec.hs
+++ b/test/NodeIPCSMSpec.hs
@@ -24,15 +24,15 @@ import           Test.Hspec (Spec, describe, it)
 
 import           Test.QuickCheck (Gen, Property, arbitrary, generate, oneof,
                                   (===))
-import           Test.QuickCheck.Monadic (assert, monadicIO, run)
+import           Test.QuickCheck.Monadic (PropertyM, assert, monadicIO, run)
 
 import           Test.StateMachine
 import           Test.StateMachine.Types
 import qualified Test.StateMachine.Types.Rank2 as Rank2
 
 import           Cardano.Shell.NodeIPC (MessageSendFailure (..), MsgIn (..),
-                                        MsgOut (..), Port (..),
-                                        ProtocolDuration (..),
+                                        MsgOut (..), NodeIPCException (..),
+                                        Port (..), ProtocolDuration (..),
                                         ServerHandles (..), clientIPCListener,
                                         closeFullDuplexAnonPipesHandles,
                                         createFullDuplexAnonPipesHandles,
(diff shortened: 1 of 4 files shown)


ksaric:
```
  test/NodeIPCSMSpec.hs:49:9: 
  1) NodeIPC state machine, Node IPC state machine, should return the correct response
       uncaught exception: IOException of type ResourceVanished
       <file descriptor: 58>: hClose: resource vanished (Broken pipe)
       (after 1 test and 7 shrinks)
         Commands
           { unCommands =
               [ Command ShutdownCmd ShutdownInitiatedResponse []
               , Command ShutdownCmd ShutdownInitiatedResponse []
               ]
           }

  test/NodeIPCSMSpec.hs:49:9: 
  1) NodeIPC state machine, Node IPC state machine, should return the correct response
       uncaught exception: IOException of type ResourceVanished
       <file descriptor: 58>: hClose: resource vanished (Broken pipe)
       (after 1 test and 5 shrinks)
         Commands
           { unCommands =
               [ Command ShutdownCmd ShutdownInitiatedResponse []
               , Command QueryPortCmd (ReplyPortResponse 12345) []
               ]
           }
```

:gun:

ksaric:
Erm, ok, this is weird, seems like I do understand how it should work - https://github.com/advancedtelematic/quickcheck-state-machine/blob/master/src/Test/StateMachine/Sequential.hs#L186

test/NodeIPCSMSpec.hs
```diff
@@ -164,27 +193,35 @@ nodeIPCSM serverHandles = StateMachine
...
+    mPostconditions _   PingCmd         _                           = Bot -- invalid
+    mPostconditions _   QueryPortCmd    (ReplyPortResponse _)       = Top -- valid
+    mPostconditions _   QueryPortCmd    _                           = Bot -- invalid
+    mPostconditions _   ShutdownCmd     (ShutdownInitiatedResponse) = Top
+    mPostconditions _   ShutdownCmd     _                           = Bot
 
     -- | Generator for symbolic actions.
     mGenerator :: Model Symbolic -> Maybe (Gen (Action Symbolic))
-    mGenerator _ =  Just $ oneof
+    mGenerator ShutdownInit = Nothing
```
ksaric: When you get `ShutdownInit`, you should stop generating any additional commands/actions?

test/NodeIPCSMSpec.hs
```diff
@@ -164,27 +193,35 @@ nodeIPCSM serverHandles = StateMachine
   where
     -- | Let's handle just Ping/Pong for now.
     mTransitions :: Model r -> Action r -> Response r -> Model r
-    mTransitions model    action      response        = model
+    mTransitions (RunningModel model) action response   = RunningModel $ model
         { inputMessages     = actionToConcrete action       : inputMessages model
         , outputMessages    = responseToConcreate response  : outputMessages model
         }
+    mTransitions _ ShutdownCmd _                        = ShutdownInit
```
ksaric: If you pick `ShutdownCmd`, it should move the model into `ShutdownInit` which should stop the generation of additional actions/commands?

test/NodeIPCSMSpec.hs
```diff
@@ -164,27 +193,35 @@ nodeIPCSM serverHandles = StateMachine
...
+    mPostconditions _   ShutdownCmd     _                           = Bot
 
     -- | Generator for symbolic actions.
     mGenerator :: Model Symbolic -> Maybe (Gen (Action Symbolic))
-    mGenerator _ =  Just $ oneof
+    mGenerator ShutdownInit = Nothing
+    mGenerator _            = Just $ oneof
         [ return PingCmd
         , return QueryPortCmd
+        --, return ShutdownCmd
```
HirotoShioi: I might be wrong but, if this is commented out, then generator will never generate `ShutdownCmd` therefore it will never test the case on Shutdown message.

HirotoShioi: But when I put it back in, it still generates commands after ShutdownCmd even though it's not supposed.

ksaric: Yes, the problem is that currently I have no clue why it doesn't work, so we can just merge it with the improved tests and new code.
If you don't uncomment it, it will work fine, just recompile it.

ksaric: There are regular tests in place, that passes.

ksaric:
# Quickcheck state machine issue

In the meantime, I did manage to run it and find the error using "regular" QuickCheck tests, but it took me more then 10000 tests to find it!
Which means, it's not an issue in `quickcheck-state-machine` but in the actual code.

I think that the reason why this is failing is that it occasionally shuts down the client and returns ExitCode, and other times we get EOF from the pipe. It's basically non-deterministic, and that is something I didn't expect.
My concern here is that it's not really acting predictable and I'm not sure if I'm supposed to "promote" an ExitCode as an actual handled exception. There are some reasons for doing that.

You can try to reproduce this using `stack test --fast -j12 --nix --test-arguments "-m NodeIPC/startIPC -a 10000"`.

Issue #183, opened 2019-06-06
Move the exception handling from node-ipc to use ExceptT


ksaric:
After some of the comments from @erikd, I concluded that what he proposed made a lot of sense and it would be a nice idea to have all the modules use the common exception handling "scheme" - https://github.com/input-output-hk/cardano-shell/pull/153#discussion_r285487166

So we should switch to `ExceptT`.

erikd:
@ksaric Maybe just run this past other members of the broader `cardano-sl` team.

CodiePP:
merged

PR #186, opened 2019-06-10
Use ExceptT on NodeIPC module

HirotoShioi:
https://github.com/input-output-hk/cardano-shell/issues/183


src/Cardano/Shell/NodeIPC/ServerExample.hs
```diff
@@ -118,11 +124,14 @@ receieveMessages serverReadHandle = do
     reply   <- readServerMessage -- Reply
     return (started, reply)
 
-getHandleFromEnv :: String -> IO Handle
+getHandleFromEnv :: String -> IO (Either NodeIPCException Handle)
```
erikd: Does `NodeIPCException` have an `Exception` instance? If so, it `Exception` instance should be dropped and the type renamed to something like `NodeIPCError`

The idea is that if an error type `e` is used as `Either e`, is should not be possible to throw it as an exception.

src/Cardano/Shell/NodeIPC/ServerExample.hs
```diff
@@ -101,13 +100,20 @@ exampleServerWithProcess msg = bracket acquire restore (action msg)
...
+    action :: MsgIn
+           -> (ReadHandle, Handle)
+           -> IO (Either NodeIPCException (MsgOut, MsgOut))
     action msgin (readHandle, _) = do
-        withCreateProcess (proc "stack" ["exec", "node-ipc", "haskell"])
-            { std_in = CreatePipe } $
-                \(Just stdIn) _ _ _ -> do
+        -- For some reason, @--no-nix@ argument is required for this function to run
+        -- currently asking DevOps regarding this.
+        withCreateProcess (proc "stack" ["exec", "--no-nix", "node-ipc", "haskell"])
```
erikd: I don't think its a good idea to make this *require* `stack` and/or `nix.

cardano-shell.cabal
```diff
@@ -118,6 +118,7 @@ executable node-ipc
       base >=4.7 && <5
     , cardano-shell
     , cardano-prelude
+    , safe-exceptions
```
ksaric: Why?

test/NodeIPCSpec.hs
```diff
@@ -74,23 +74,23 @@ nodeIPCSpec = do
                     assert $ parseError == (MessageOutFailure $ ParseError errorMessage)
 
             it "should throw NodeIPCException when closed handle is given" $ monadicIO $ do
-                eResult <- run $ try $ do
+                eResult <- run $ do
```
ksaric: When tests simplify, it usually means we are getting closer to cleaner code.

src/Cardano/Shell/NodeIPC/Lib.hs
```diff
@@ -200,35 +201,69 @@ instance Show NodeIPCException where
...
         Just fdstring -> case readEither fdstring of
-            Left err -> throwM $ UnableToParseNodeChannel $ toS err
-            Right fd -> liftIO $ fdToHandle fd
+            Left err -> throwError $ UnableToParseNodeChannel $ toS err
+            Right fd -> liftIO (fdToHandle fd)
 
 -- | Start IPC with given 'ReadHandle', 'WriteHandle' and 'Port'
-startIPC :: forall m. (MonadIO m) => ProtocolDuration -> ReadHandle -> WriteHandle -> Port -> m ()
-startIPC protocolDuration readHandle writeHandle port = liftIO $ void $ ipcListener protocolDuration readHandle writeHandle port
+startIPC :: forall m. (MonadIO m, MonadMask m)
```
ksaric: `hindent`?
Manual formatting is much cleaner.

src/Cardano/Shell/NodeIPC/Lib.hs
```diff
@@ -200,35 +201,69 @@ instance Show NodeIPCException where
...
+         -> m (Either NodeIPCException ())
+startIPC protocolDuration readHandle writeHandle port =
+    runNodeIPC $ ipcListener protocolDuration readHandle writeHandle port
 
 -- | Start IPC with NodeJS
 --
 -- This only works if NodeJS spawns the Haskell executable as child process
 -- (See @server.js@ as an example)
-startNodeJsIPC :: forall m. (MonadIO m) => ProtocolDuration -> Port -> m ()
+startNodeJsIPC :: (MonadIO m)
```
ksaric: `hindent`?
Manual formatting is much cleaner.

src/Cardano/Shell/NodeIPC/ServerExample.hs
```diff
@@ -118,11 +124,14 @@ receieveMessages serverReadHandle = do
...
     case mFdstring of
-        Nothing -> throwM $ NodeChannelNotFound (strConv Lenient envName)
+        Nothing -> throw $ NodeChannelNotFound (strConv Lenient envName)
         Just fdstring -> case readEither fdstring of
-            Left err -> throwM $ UnableToParseNodeChannel (strConv Lenient err)
-            Right fd -> liftIO $ fdToHandle fd
+            Left err -> throw $ UnableToParseNodeChannel (strConv Lenient err)
+            Right fd -> Right <$> fdToHandle fd
+  where
+    throw :: NodeIPCException -> IO (Either NodeIPCException Handle)
```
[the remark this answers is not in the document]
erikd: I didn't read that in context which is why I made the comment I made. I think it should be called just about anything other than `throw`, because that implies exceptions.

src/Cardano/Shell/NodeIPC/ServerExample.hs
erikd: If `NodeIPCException` does not have an `Exception` instance (or if it us removed), it should should be renamed to something like `NodeIPCError`

HirotoShioi: Ah okay

src/Cardano/Shell/NodeIPC/Lib.hs
```diff
@@ -200,35 +201,69 @@ instance Show NodeIPCException where
...
+--------------------------------------------------------------------------------
+
+-- | Monad stack for node-ipc
+newtype NodeIPCMonad m a = NodeIPCMonad
+    { getNodeIPC :: ExceptT NodeIPCException m a
+    } deriving ( Functor
+               , Applicative
+               , Monad
+               , MonadIO
+               , MonadError NodeIPCException
```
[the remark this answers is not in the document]
HirotoShioi: If I remove either `ExceptT` or `MonadError`, ghc starts complaining. Do you know if there's any workaround?

src/Cardano/Shell/NodeIPC/Lib.hs
```diff
@@ -244,15 +279,19 @@ handleIPCProtocol (Port port) = \case
...
     => ProtocolDuration
     -> ReadHandle
     -> WriteHandle
     -> Port
     -> m ()
-ipcListener protocolDuration readHandle@(ReadHandle rHndl) writeHandle@(WriteHandle wHndl) port = do
-    checkHandles readHandle writeHandle
-    handleMsgIn `catches` [Handler handler, Handler handleMsgError] `finally` shutdown
+ipcListener protocolDuration readHandle@(ReadHandle rHndl) writeHandle@(WriteHandle wHndl) port =
+    ( do
```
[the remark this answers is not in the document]
ksaric: Extract the function into a separate function in a where condition?

src/Cardano/Shell/NodeIPC/Lib.hs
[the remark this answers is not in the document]
erikd: `ExceptT` and `MonadError` are basically the same thing so it should be possible to remove one of them (preferably `MonadError`).

Ping me on slack if you are still having trouble.

cardano-shell.cabal
CodiePP: should use everywhere functions from `Control.Exception.Safe`?

src/Cardano/Shell/NodeIPC/ServerExample.hs
CodiePP: to run the example from the source directory this looks like a reasonable way to do it.

cardano-shell.cabal
erikd: The purpose of [`Control.Exception.Safe`](https://hackage.haskell.org/package/safe-exceptions) is to avoid some abuses of the `SomeException` type (like catching catching all exceptions and either silently swallowing them, and/or retrying the the function that threw it).

In my opinion this is a poor alternative to:
* Handling all exceptions correctly.
* Avoiding the use of exceptions for control flow and error handing (ie using `Either` and `ExceptT` instead).
* Catching exceptions thrown by upstream libraries and converting them to error types.

app/NodeIPCClient/Main.hs
```diff
@@ -28,27 +28,24 @@ port = Port 8090
...
-    restore = hClose
-
-    action :: Handle -> IO ()
+    action :: Handle -> IO (Either NodeIPCError ())
     action serverWHandle = do
+        hSetBuffering serverWHandle LineBuffering
         let serverWriteHandle = WriteHandle serverWHandle
         let serverReadHandle  = ReadHandle stdin
         startIPC SingleMessage serverReadHandle serverWriteHandle port
+        `finally`
```
ksaric: Indent it, please, it should be visible it's connected to the previous command.

app/NodeIPCClient/Main.hs
```diff
@@ -28,27 +28,24 @@ port = Port 8090
...
     (cmd:_) <- getArgs
     case cmd of
-        "js"      -> startNodeJsIPC SingleMessage port
-        "haskell" -> bracket acquire restore action
-        _         -> return ()
+        "js"      -> Right <$> void (startNodeJsIPC SingleMessage port)
+        "haskell" -> either
+            (return . Left)
+            action
+            =<<
```
ksaric: Can you make this a bit more readable?

src/Cardano/Shell/NodeIPC/Lib.hs
```diff
@@ -264,55 +278,56 @@ ipcListener protocolDuration readHandle@(ReadHandle rHndl) writeHandle@(WriteHan
...
             sendMessage writeHandle messageByteString            -- Write to client/server
 
-    send :: MsgOut -> m ()
+    send :: MsgOut -> ExceptT NodeIPCError IO ()
     send = sendMessage writeHandle
 
-    handler :: IOError -> m ()
+    handler :: IOError -> ExceptT NodeIPCError IO ()
     handler err = do
         liftIO $ when (isEOFError err) $ logError "its an eof"
```
ksaric: This line is a bit confusing. We can talk about that in the next PR.

src/Cardano/Shell/NodeIPC/ServerExample.hs
cleverca22: also, using `stack exec node-ipc` breaks the ability to pass file-handles to the child process, which is why the `js` vs `haskell` flag exists here

`$(stack path --local-install-root)/bin/node-ipc` will return the path to the `node-ipc` binary made by stack, allowing you to run it in `js` mode

so the only missing part, is to either ensure the `ServerExample.hs` is always ran with `$(stack path --local-install-root)/bin` added to `PATH` (then you can just assume `node-ipc` is a valid cmd)
or you need to try `node-ipc` and then fallback to asking stack where it was

ksaric:
Bors responding ...

ksaric:
Bors complete.

ksaric:
We are the bors...

PR #187, opened 2019-06-13
add automated testing on posix platforms

cleverca22:


HirotoShioi:
Since you're deleting `cleanup` function. You should delete this line.

https://github.com/input-output-hk/cardano-shell/blob/0f5950aca2bf0b67fa590612bc51ad7e4bd4465a/app/NodeIPCClient/server.js#L28

PR #190, opened 2019-06-19
adapt startIPC so it can take any fromjson/tojson message pair

cleverca22:
this PR moves the handler for the IPC into `Cardano.Shell.NodeIPC.SimplePortServer` along with the types for the protocol

and then adapts all related code to take a handler and generic `FromJSON` and `ToJSON` types, so that another program can just provide its own replacement of `Cardano.Shell.NodeIPC.SimplePortServer` and reuse the bulk of the logic

todo:
- [ ] rebase once https://github.com/input-output-hk/cardano-shell/pull/186 is merged
- [ ] clean up imports and stylish-haskell
- [ ] replace `stack exec node-ipc` with `$(stack path --local-install-root)/bin/node-ipc` so the haskell mode with stdin/out isnt needed

### DIFF
--- a/app/NodeIPCClient/Main.hs
+++ b/app/NodeIPCClient/Main.hs
@@ -12,7 +12,7 @@ import           Cardano.Prelude
 import           Cardano.Shell.NodeIPC (Port (..), ProtocolDuration (..),
                                         ReadHandle (..), WriteHandle (..),
                                         getHandleFromEnv, startIPC,
-                                        startNodeJsIPC)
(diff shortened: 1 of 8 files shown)


ksaric:
@cleverca22 We will take a look at what to do on Monday and will respond with a comment/review then.

ksaric:
@cleverca22 We are going to use the existing implementation from the `wallet-backend`, the original code is going to remain the same. Closing.

PR #200, opened 2019-06-28
Small typo

jamiegabbay:
an thin layer
->
a thin layer


ksaric:
@jamiegabbay I think you need to be part of the https://github.com/input-output-hk organization to trigger the builds (in any project).

PR #204, opened 2019-07-01
Add pkgconfig and systemd to nix-shell for stack.

jbgi:
Also build shell on osx on hydra.

This fix https://buildkite.com/input-output-hk/cardano-shell/builds/688#61197167-e9f5-4953-9033-8815a06a9a2e

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -55,12 +55,12 @@ let
 in defaultNix // {
   # ... and add additional packages we want to build on CI:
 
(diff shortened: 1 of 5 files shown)


erikd:
Ah crap, I did not realise that this PR was against my branch. I thought it was against `master`.

Have now pulled that into my branch. Lets hope it works :crossed_fingers:

Issue #208, opened 2019-07-02
Add node-ipc for wallet


ksaric:
During the discussions with the wallet-backend, we decided to use their implementation for node-ipc since they say they have a specific way of handling the logic of the node-ipc.

We will leave the old implementation in the `NodeIPC`, which has been exhaustively tested and we will have their custom implementation in `DaedalusIPC`, which only has integration tests.

## More details

It is a mostly identical version of what we did for several months, with very little difference.

To be clear, I do not agree with that decision but in the interest of time and stability, we will simply copy their implementation so they can use it. 

_We are not responsible for any bugs or errors coming from that implementation. They said they want it that way, they can have it and be responsible adults about it._

There are several reasons why I don't agree with that "replacement", the crucial difference being the exhaustive test coverage, convoluted code, the closing of the client when the pipe is closed and no totality checks.

But since it seems that they seem to know what they are doing, they can use their own implementation.

PR #209, opened 2019-07-02
[GH-208] Add node-ipc for wallet.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/208

termed "DaedalusIPC"


src/Cardano/Shell/DaedalusIPC.hs
```diff
@@ -0,0 +1,233 @@
...
+import qualified Data.ByteString.Lazy as BL
+import qualified Data.ByteString.Lazy.Char8 as L8
+import qualified Data.Text as T
+
+----------------------------------------------------------------------------
+-- Daedalus <-> Wallet child process port discovery protocol
+
+data MsgIn = QueryPort
+    deriving (Show, Eq)
+data MsgOut = Started | ReplyPort Int | ParseError Text
```
rvl: We reduce these even further. I think the 4 messages could just be one:
```haskell
data MsgOut = ListenPort Int deriving (Show, Eq)
```
Daedalus would need to be modified though.

src/Cardano/Shell/DaedalusIPC.hs
```diff
@@ -0,0 +1,233 @@
...
+    msg (Right QueryPort) = Just (ReplyPort port)
+    msg (Left e)          = Just (ParseError e)
+
+    -- What to do in context of withNodeChannel
+    action :: (MsgOut -> IO ()) -> IO ()
+    action send = send Started >> sleep
+
+    sleep = threadDelay maxBound
+
+----------------------------------------------------------------------------
```
rvl: Everything below this line is not specific to Daedalus and is just for the nodejs ipc protocol. It could go in a separate module.

src/Cardano/Shell/DaedalusIPC.hs
```diff
@@ -0,0 +1,233 @@
...
+    -- How to respond to an incoming message, or when there is an incoming
+    -- message that couldn't be parsed.
+    msg (Right QueryPort) = Just (ReplyPort port)
+    msg (Left e)          = Just (ParseError e)
+
+    -- What to do in context of withNodeChannel
+    action :: (MsgOut -> IO ()) -> IO ()
+    action send = send Started >> sleep
+
+    sleep = threadDelay maxBound
```
rvl: I think this needs to be:
```suggestion
    sleep = forever $ threadDelay maxBound
```

src/Cardano/Shell/DaedalusIPC.hs
```diff
@@ -0,0 +1,233 @@
...
+-- reading or writing to its file descriptor.
+newtype NodeChannelFinished = NodeChannelFinished IOError
+
+-- | Communicate with a parent process using a NodeJS-specific protocol. This
+-- process must have been spawned with one of @stdio@ array entries set to
+-- @'ipc'@.
+--
+-- If the channel could be set up, then it returns a function for communicating
+-- with the parent process.
+withNodeChannel
```
rvl: You are right. This function signature is way too complicated. How about this:

```haskell
newtype NodeChannel = NodeChannel Handle

setupNodeChannel :: IO (Either NodeChannelError NodeChannel)

runNodeChannel
    :: (FromJSON msgin, ToJSON msgout)
    => (Either Text msgin -> IO (Maybe msgout))
       -- ^ Handler for messages coming from the parent process. Left values are
       -- for JSON parse errors. The handler can optionally return a reply
       -- message.
    -> ((msgout -> IO ()) -> IO a)
       -- ^ Action to run with the channel. It is passed a function for sending
       -- messages to the parent process.
    -> NodeChannel
       -- ^ Channel provided by 'setupNodeChannel'
    -> IO (Either NodeChannelFinished a)
```

These two functions and the `NodeChannel` type, but not its constructor would be exported from the module.

src/Cardano/Shell/DaedalusIPC.hs
```diff
@@ -0,0 +1,233 @@
...
+    maybeSend :: Maybe msgout -> IO ()
+    maybeSend = maybe (pure ()) (putMVar chan)
+
+    unexpected = userError "ipcListener: unreachable code"
+
+readMessage :: Handle -> IO BL.ByteString
+readMessage = if isWindows then windowsReadMessage else posixReadMessage
+
+isWindows :: Bool
+isWindows = arch == "windows"
```
rvl: This should be `System.Info.os` (doh).

src/Cardano/Shell/DaedalusIPC.hs
```diff
@@ -0,0 +1,233 @@
...
+
+sendMessage :: Handle -> BL.ByteString -> IO ()
+sendMessage handle msg = send handle msg >> hFlush handle
+  where
+    send = if isWindows then sendMessageWindows else sendMessagePosix
+
+sendMessageWindows :: Handle -> BL.ByteString -> IO ()
+sendMessageWindows = sendWindowsMessage' 1 0
+
+sendWindowsMessage' :: Word32 -> Word32 -> Handle -> BL.ByteString -> IO ()
```
rvl: Naming's inconsistent. This should be `sendMessageWindows'`

src/Cardano/Shell/DaedalusIPC.hs
```diff
@@ -0,0 +1,233 @@
...
+    sendMsg :: msgout -> IO ()
+    sendMsg = sendMessage handle . encode
+
+    maybeSend :: Maybe msgout -> IO ()
+    maybeSend = maybe (pure ()) (putMVar chan)
+
+    unexpected = userError "ipcListener: unreachable code"
+
+readMessage :: Handle -> IO BL.ByteString
+readMessage = if isWindows then windowsReadMessage else posixReadMessage
```
rvl: Naming's inconsistent. These should be `readMessageWindows` and `readMessagePosix`.

CodiePP:
this PR is now rotting for almost two weeks! and it's getting complicated with the proposed changes by @rvl being in another PR against this one. Could this be sped up a bit, please!

rvl:
I have merged in the clean ups from #213. The PR #221 adds a nodejs test case.

rvl:
This is ready to merge. @ksaric @CodiePP 
The Hydra failures are due to broken MacOS builders. The `tests.ipc.x86_64-darwin` job has already been [failing on master](https://hydra.iohk.io/job/Cardano/cardano-shell/tests.ipc.x86_64-darwin), so the failure is not caused by this code change.

PR #212, opened 2019-07-03
updating logging to the latest version

CodiePP:
Update to the latest version of `iohk-monitoring-framework` with a breaking change!

* function `appendName` is now pure

----

it turns out that we have to update each project that integrates with `cardano-shell` to get this updated logging features into `cardano-node` :-)


app/Cardano/Shell/Features/Logging.hs
```diff
@@ -71,7 +71,7 @@ data LoggingLayer = LoggingLayer
     , llLogNotice       :: forall m a.      (MonadIO m, Show a)  => Trace m a  -> a -> m ()
     , llLogWarning      :: forall m a.      (MonadIO m, Show a)  => Trace m a  -> a -> m ()
     , llLogError        :: forall m a.      (MonadIO m, Show a)  => Trace m a  -> a -> m ()
-    , llAppendName      :: forall m a.      (MonadIO m, Show a)  => LoggerName -> Trace m a -> m (Trace m a)
+    , llAppendName      :: forall m a.      (MonadIO m, Show a)  => LoggerName -> Trace m a -> Trace m a
```
365andreas: MonadIO is not necessary here.

CodiePP: yes, right!

PR #214, opened 2019-07-05
Added Configuration in LoggingLayer.

365andreas:
provides access to logging `Configuration`

Convenience! this is needed in cardano-node to flexibly work on logging, benchmarking and monitoring. Once this has been finished we might rethink on how to export the required functions through the logging layer in the shell.


app/Cardano/Shell/Features/Logging.hs
```diff
@@ -152,6 +153,7 @@ loggingCardanoFeatureInit loggingConfiguration = do
                     , llBracketMonadX   = Monadic.bracketObserveX logConfig
                     , llBracketStmIO    = Stm.bracketObserveIO logConfig
                     , llBracketStmLogIO = Stm.bracketObserveLogIO logConfig
+                    , llConfiguration   = lpConfiguration loggingCfg
```
CodiePP: should be `logConfig` here

365andreas: Done.

PR #215, opened 2019-07-08
allow integration with user-defined backends for logging

CodiePP:
add integration of user-defined backends to logging

`LoggingLayer` now exports this function:

```
llAddBackend      :: Backend Text -> Text -> IO ()
```

### DIFF
--- a/app/Cardano/Shell/Features/Logging.hs
+++ b/app/Cardano/Shell/Features/Logging.hs
@@ -24,6 +24,8 @@ import           Control.Exception.Safe (MonadCatch)
 import qualified Control.Monad.STM as STM
 import           Options.Applicative
 
+import           Cardano.BM.Data.Backend (Backend)
+import qualified Cardano.BM.Backend.Switchboard as Switchboard
 import           Cardano.BM.Configuration (Configuration)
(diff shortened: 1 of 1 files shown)


app/Cardano/Shell/Features/Logging.hs
```diff
@@ -78,6 +80,7 @@ data LoggingLayer = LoggingLayer
     , llBracketStmIO    :: forall a t.      (Show a) => Trace IO a -> Severity -> Text -> STM.STM t -> IO t
     , llBracketStmLogIO :: forall a t.      (Show a) => Trace IO a -> Severity -> Text -> STM.STM (t,[(LOMeta, LOContent a)]) -> IO t
     , llConfiguration   :: Configuration
+    , llAddBackend      :: Backend Text -> Text -> IO ()
```
ksaric: `Either ...` with the error being the error that can occur when a user selects the SwitchBoard multiple times or something like that?

CodiePP: the user does not have access to the switchboard, so cannot connect it again.
but, yes this would be uncomfortable. we will see if other error cases can occur.

Issue #210, opened 2019-07-02
Configuration machinery integration


deepfire:
# Context

Users of the configuration API would benefit from having a single source of truth for configurable values.  The status quo in `cardano-shell` has three different configuration mechanisms which are not seemingly integrated with each other, at least in context of `cardano-node`:

1. CLI (see https://github.com/input-output-hk/cardano-node/blob/master/app/Main.hs#L119)
2. Dhall config files (no mention in `cardano-node`)
3. Presets (`CardanoConfiguration` in https://github.com/input-output-hk/cardano-shell/blob/master/src/Cardano/Shell/Presets.hs#L145)

**Note**: *this is a prerequisite for* https://github.com/input-output-hk/cardano-node/issues/30.

# Requirements

1. Configuration API users should have a single, unified, final source of truth for configuration values, that is informed by the aforementioned mechanisms.
2. Maintenance of the separation between relative "constants" and "configurables". Sometimes the distinction is hard to make.
   - Here _constants_ are defined as environment/cluster defaults that are not necessarily available to CLI-based overriding.
3. No hard coded values: there should be no need to recompile `cardano-node` to override any values, even those that are classified as "constants". This does not mean that all values need to be necessarily configurable via CLI -- any runtime configuration specification method is acceptable.

# Open questions

1. There is no overarching type for the overall configuration.  Do we want one?  To what degree do we think we want/can keep configuration isolated per feature?  How do we approach ambiguities?
2. What is the authoritative definition for the notion of "constants"?  Are they indeed just a combination of two properties:
    a. Having environment (mainnet/testnet/staging/dev) -dependent defaults.
    b. Not having CLI representation.

    Do we, perhaps, want to separate these two properties and apply them to each option independently?

# Ideas

1. Towards the goal of a unified model of configuration, we could assemble a single registry of all configuration options.  That could help with:
   - avoiding duplication
   - finding the best place for new options
   - keeping track of the properties of the options -- users, environment dependency, CLI-overridability
   - maintaining an overall picture
   - setting up a point of collaboration in this area
  One could imagine this registry being kept as a spreadsheet -- since that would allow sorting on arbitrary columns, so as to allow easy queries.

2. Towards the requirement №2, we could externalise the hard-coded values for the `CardanoConfiguration` values into a external file -- and a Dhall file, for that matter, so as to reuse an existing mechanism.

[...]

PR #217, opened 2019-07-10
[GH-210] Initial configuration machinery integration.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/210

Initial configuration machinery integration.
This is working just for the **Core** type by using constants and CLI, but shows how it can work with other configuration types as well.

https://medium.com/@jonathangfischoff/the-partial-options-monoid-pattern-31914a71fc67

# Examples

```
stack exec cardano-shell-exe --nix

Missing: --log-config LOGCONFIG

Usage: cardano-shell-exe --log-config LOGCONFIG [--src-file-path SRC-FILE-PATH]
                         [--genesis-hash GENESIS-HASH]
                         [--prev-block-hash PREV-BLOCK-HASH]
                         ([--require-network-magic] |
                         [--no-require-network-magic]) [--db-version DB-VERSION]
  Logging Feature
```

```
stack exec cardano-shell-exe --nix -- --log-config configuration/log-configuration.yaml 

The DB versionLast {getLast = Just (Last {getLast = Nothing})}
{"at":"2019-07-10T14:24:09.08Z","env":"fromList [(\"value\",String \"Release-1.0.0\")]:0.1.10.0","ns":["iohk","cardano"],"data":{},"app":["iohk"],"msg":"Hello from logging layer ...","pid":"27076","loc":null,"host":"hostname","sev":"Notice","thread":"ThreadId 15"}
{"at":"2019-07-10T14:24:09.08Z","env":"fromList [(\"value\",String \"Release-1.0.0\")]:0.1.10.0","ns":["iohk","cardano","cardano-shell"],"data":{},"app":["iohk"],"msg":"Hello #2 from logging layer ...","pid":"27076","loc":null,"host":"hostname","sev":"Notice","thread":"ThreadId 15"}
Running node/wallet/whatever!
Running node/wallet/whatever!
Running node/wallet/whatever!
Running node/wallet/whatever!
Running node/wallet/whatever!
cardano-shell-exe: Something went wrong and we don't know what.
```

```
stack exec cardano-shell-exe --nix -- --log-config configuration/log-configuration.yaml --db-version 12345

The DB versionLast {getLast = Just (Last {getLast = Just 12345})}
{"at":"2019-07-10T14:25:00.52Z","env":"fromList [(\"value\",String \"Release-1.0.0\")]:0.1.10.0","ns":["iohk","cardano"],"data":{},"app":["iohk"],"msg":"Hello from logging layer ...","pid":"31401","loc":null,"host":"hostname","sev":"Notice","thread":"ThreadId 15"}
{"at":"2019-07-10T14:25:00.52Z","env":"fromList [(\"value\",String \"Release-1.0.0\")]:0.1.10.0","ns":["iohk","cardano","cardano-shell"],"data":{},"app":["iohk"],"msg":"Hello #2 from logging layer ...","pid":"31401","loc":null,"host":"hostname","sev":"Notice","thread":"ThreadId 15"}
Running node/wallet/whatever!
Running node/wallet/whatever!
Running node/wallet/whatever!
Running node/wallet/whatever!
Running node/wallet/whatever!
cardano-shell-exe: Something went wrong and we don't know what.
```

### Do we need to use lenses at this point?

I usually don't want them unless there is a lot of manipulation of the data structures we need to think about, since it has a cost. Nothing is free.
At this point, we could include them, but it might include some dependencies and add some complexity to the code.

http://hackage.haskell.org/package/microlens

### Do we need to add the configuration file options right now or wait until we really need them?

I still want to see what we need before I just randomly create something.
If it turns out we don't need config files, we might not use them, but I would like to see a use case first.


app/Cardano/Shell/Features/Networking.hs
```diff
@@ -75,6 +76,8 @@ createNetworkingFeature loggingLayer cardanoEnvironment cardanoConfiguration = d
     -- the filesystem, so we give him the most flexible/powerful context, @IO@.
     networkingConfiguration <-  pure "THIS IS AN EXAMPLE OF A CONFIGURATION!"
 
+    putTextLn $ "The DB version" <> (show $ coDBSerializeVersion <$> ccCore cardanoConfiguration)
```
ksaric: Just an example so we can see that it actually works.

app/Main.hs
```diff
@@ -77,12 +86,30 @@ initializeAllFeatures :: CardanoConfiguration -> CardanoEnvironment -> IO ([Card
 initializeAllFeatures cardanoConfiguration cardanoEnvironment = do
 
     -- Here we parse the __CLI__ arguments for the actual application.
-    loggingCLIArguments             <- execParser parserWithInfo
+    CLIArguments loggingCLIArguments coreConfig <- execParser parserWithInfo
+
+    -- TODO(KS): Looks like we might need to include lenses at some point?
+    let cardanoConfiguration' = CardanoConfiguration
```
deepfire: Why not:
```haskell
cardanoConfiguration' = cardanoConfiguration { ccCore = ccCore cardanoConfiguration <> pure coreConfig }
```
?

ksaric: Fixed.

PR #219, opened 2019-07-11
 Next step in the partial-options-monoid implementation

deepfire:
Take the next step in the partial-options-monoid implementation:

1. Finalisation functions that handle partiality
2. Cover `CardanoConfiguration`
3. Reverts to having unchanged `CardanoConfiguration` and `Core` types, using the `Last` type only in the `Partial`* variants.

Part of https://github.com/input-output-hk/cardano-shell/issues/210

### DIFF
--- a/src/Cardano/Shell/Configuration/Lib.hs
+++ b/src/Cardano/Shell/Configuration/Lib.hs
@@ -42,41 +42,43 @@ lastToEither errMsg (Last x) = maybe (Left errMsg) Right x
 -- in the partial options monoid approach, after all the parametrisation layers have been merged,
 -- and we're intending to use the resultant config -- they ensure that all values are defined.
 --
+-- NOTE: we should look into applying generic programming and/or TH for this boilerlate.
+--
 finaliseCardanoConfiguration :: PartialCardanoConfiguration -> Either Text CardanoConfiguration
 finaliseCardanoConfiguration PartialCardanoConfiguration{..} = do
-  ccLogPath                <- lastToEither "Unspecified ccLogPath"      pccLogPath
-  ccLogConfig              <- lastToEither "Unspecified ccLogConfig"    pccLogConfig
-  ccDBPath                 <- lastToEither "Unspecified ccDBPath"       pccDBPath
-  ccApplicationLockFile    <- lastToEither "Unspecified ccApplicationLockFile"
-                                                                        pccApplicationLockFile
-  ccCore                   <- join $ finaliseCore <$>
-                              lastToEither "Unspecified ccCore"         pccCore
-  ccNTP                    <- lastToEither "Unspecified ccNTP"          pccNTP
-  ccUpdate                 <- lastToEither "Unspecified ccUpdate"       pccUpdate
-  ccTXP                    <- lastToEither "Unspecified ccTXP"          pccTXP
-  ccSSC                    <- lastToEither "Unspecified ccSSC"          pccSSC
-  ccDLG                    <- lastToEither "Unspecified ccDLG"          pccDLG
-  ccBlock                  <- lastToEither "Unspecified ccBlock"        pccBlock
-  ccNode                   <- lastToEither "Unspecified ccNode"         pccNode
-  ccTLS                    <- lastToEither "Unspecified ccTLS"          pccTLS
-  ccWallet                 <- lastToEither "Unspecified ccWallet"       pccWallet
-
-  pure CardanoConfiguration{..}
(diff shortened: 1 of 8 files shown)


src/Cardano/Shell/Configuration/Lib.hs
```diff
@@ -23,6 +34,50 @@ import           Cardano.Shell.Configuration.Types (BlockchainConfig,
...
+lastToEither errMsg (Last x) = maybe (Left errMsg) Right x
+
+--
+-- The finalise* family of functions are supposed to be called at the very last stage
+-- in the partial options monoid approach, after all the parametrisation layers have been merged,
+-- and we're intending to use the resultant config -- they ensure that all values are defined.
+--
+finaliseCardanoConfiguration :: PartialCardanoConfiguration -> Either Text CardanoConfiguration
+finaliseCardanoConfiguration PartialCardanoConfiguration{..} = do
+  ccLogPath                <- lastToEither "Unspecified ccLogPath"      pccLogPath
```
ksaric: We use 4 spaces in this repo.

src/Cardano/Shell/Constants/CLI.hs
```diff
@@ -19,13 +19,16 @@ import           Cardano.Shell.Constants.Types
 lastOption :: Parser a -> Parser (Last a)
 lastOption parser = Last <$> optional parser
 
+lastStrOption :: IsString a => Mod OptionFields a -> Parser (Last a)
```
ksaric: My eyes :eye: 
Do we really need such micro improvements?
They sometimes introduce more complexity at the price of a little duplication.

Some duplication is generally ok.

src/Cardano/Shell/Lib.hs
```diff
@@ -147,7 +149,7 @@ runCardanoApplicationWithFeatures _ cardanoFeatures cardanoApplication = do
         pure ()
 
 
-type AllFeaturesInitFunction = CardanoConfiguration -> CardanoEnvironment -> IO [CardanoFeature]
+type AllFeaturesInitFunction = PartialCardanoConfiguration -> CardanoEnvironment -> IO [CardanoFeature]
```
ksaric: This is kind of weird. What is the point of `CardanoConfiguration` then?

src/Cardano/Shell/Presets.hs
```diff
@@ -20,25 +22,25 @@ import           Cardano.Shell.Constants.Types (Block (..),
...
-    , ccDBPath              = "./db/"
-    , ccApplicationLockFile = ""
-    , ccCore =
-        pure Core
-          { coGenesis =
-              pure Genesis
-                { geSrc             = "mainnet-genesis.json"
-                , geGenesisHash     = "89d9b5a5b8ddc8d7e5a6795e9774d97faf1efea59b2caf7eaf9f8c5b32059df4"
-                , gePrevBlockHash   = "5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb"
+  PartialCardanoConfiguration
```
ksaric: More weirdness. 
What is the point of separation of `CardanoConfiguration` and `PartialCardanoConfiguration` if there is no point to it :smile:

src/Cardano/Shell/Lib.hs
deepfire: The idea is that `Partial*` variants are:

- allowed to be incomplete
- mergeable

So the `finalise*` family functions serve as a decision point where we say "ok, we merged all sources of configuration, and the the result is fully defined and ready for use".

src/Cardano/Shell/Constants/CLI.hs
deepfire: The idea is that _all_ the `strOption`'s will end up needing a `Last` wrapper -- that's in the nature of the approach.  Unless I'm missing something, of course..

src/Cardano/Shell/Presets.hs
deepfire: I guess what is missing here is the picture of how we arrive at the final value of `CardanoConfiguration`, in a generic way.  This goes back to the call we had between the three of us with @CodiePP.

I do agree that in the _current_ form, there's no use for the specific `PartialCardanoConfiguration`, since we're not merging them yet.  But this gives us an opportunity for transparent merging of feature-provided `PartialCardanoConfiguration`'s.

So the short answer is -- it's there to provide mergeability.

src/Cardano/Shell/Configuration/Lib.hs
deepfire: Apologies, fixed!

PR #221, opened 2019-07-13
Add DaedalusIPC nodejs test

rvl:
For PR #209. Adds a test case which uses nodejs to launch DaedalusIPC.


rvl:
I'm not sure why that test might fail like that now... any ideas @cleverca22 ?

Issue #218, opened 2019-07-11
Add Node specific configuration options using CLI


ksaric:
Continuation from https://github.com/input-output-hk/cardano-shell/pull/217 for the Node specific part of the configuration.

PR #222, opened 2019-07-15
[GH-218] Add Node specific configuration options using CLI.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/218


deepfire:
On the other hand, having a separate file for the `Partial` types is a good hint at what exactly we should derive/generate automatically.

Ideally we should get this functionality with no structural duplication / "for free".

Issue #216, opened 2019-07-08
Add the update installation to the launcher


ksaric:
Since for now, the launcher is going to be in charge of the installation of the updates, we can add that in the launcher or `cardano-shell`.

ksaric:
@cleverca22 Can you help us out here with the Nix integration so we can have the update system calls?

https://github.com/input-output-hk/cardano-shell/wiki/Update-mechanism

ksaric:
This is where the Windows installer _cross-compilation_ Nix expression is - https://github.com/input-output-hk/daedalus/blob/develop/default.nix#L154

We have several ways of running the update and it depends on which platform we are, but in the end we just run the installer again.
For example, this is a specific case for Windows - https://github.com/input-output-hk/cardano-sl/blob/develop/tools/src/launcher/Main.hs#L591-L610

We have similar cases for Mac and Linux, but we can hardcode them in the Haskell code as well.
We can make it as simple as possible.

ksaric:
For the cases of Mac and Linux, we can see the DHall config we use for those platforms.

Mac:
- https://github.com/input-output-hk/daedalus/blob/develop/installers/dhall/macos64.dhall#L31
- https://github.com/input-output-hk/daedalus/blob/develop/installers/dhall/macos64.dhall#L33

Linux:
- https://github.com/input-output-hk/daedalus/blob/develop/installers/dhall/linux64.dhall#L29
- https://github.com/input-output-hk/daedalus/blob/develop/installers/dhall/linux64.dhall#L31

ksaric:
For testing, we can use a new version of the application by using the Nix build - https://github.com/input-output-hk/daedalus/blob/develop/default.nix#L12

ksaric:
This is _optional_ to take a look at, since it might get a bit complicated (ignore it at first, we can talk about it when we have the thing running), but @cleverca22 has already developed the tool for update proposals which can be seen here - https://github.com/input-output-hk/cardano-sl/tree/develop/script-runner

ksaric:
This is from where the installer gets the updates - https://github.com/input-output-hk/daedalus/blob/develop/installers/dhall/mainnet.dhall#L4

PR #231, opened 2019-07-19
Hiroto/update system

HirotoShioi:
https://github.com/input-output-hk/cardano-shell/issues/216
WIP


src/Cardano/Shell/Update/Lib.hs
```diff
@@ -0,0 +1,108 @@
...
+                Just windowsPath -> do
+                    writeWindowsUpdaterRunner windowsPath
+                    runCmd windowsPath args archive
+            case exitCode of
+                ExitSuccess -> do
+                    whenJust (udArchivePath ud) $ \updateArchivePath -> do
+                        removeFile updateArchivePath
+                    return . Right $ ()
+                ExitFailure code -> return . Left $ UpdateFailed code
+        else return . Left $ UpdaterDoesNotExist
```
ksaric: Newline after `else`.

src/Cardano/Shell/Update/Lib.hs
```diff
@@ -0,0 +1,108 @@
...
+runUpdater :: UpdaterData -> IO (Either UpdateError ())
+runUpdater ud = do
+    let path = udPath ud
+    let args = map toS $ udArgs ud
+    let mWindowPath = udWindowsPath ud
+    let archive = maybe [] (\arch -> [toS arch]) (udArchivePath ud)
+    updaterExists <- doesFileExist path
+    if updaterExists
+        then do
+            exitCode <- case mWindowPath of
```
ksaric: I don't think it should be optional.
An actual path should be there so it shouldn't be `Maybe FilePath` but an actual `FilePath` for an executable for each platform.

src/Cardano/Shell/Update/Lib.hs
```diff
@@ -0,0 +1,108 @@
...
+                    runCmd windowsPath args archive
+            case exitCode of
+                ExitSuccess -> do
+                    whenJust (udArchivePath ud) $ \updateArchivePath -> do
+                        removeFile updateArchivePath
+                    return . Right $ ()
+                ExitFailure code -> return . Left $ UpdateFailed code
+        else return . Left $ UpdaterDoesNotExist
+  where
+    runCmd path args archive =
```
ksaric: Please add the types to an anonymous (or where/let) function when you are writing it to make it more clear.

src/Cardano/Shell/Update/Lib.hs
ksaric: Also, is it possible to do `return . Left . UpdaterDoesNotExist`?
Just wondering.

src/Cardano/Shell/Update/Lib.hs
```diff
@@ -0,0 +1,108 @@
...
+        withCreateProcess (proc path (args <> archive))
+            -- WIP
+            $ \_in _out _err ph -> waitForProcess ph
+            
+    whenJust (Just a) f = f a
+    whenJust Nothing  _ = return ()
+
+writeWindowsUpdaterRunner :: FilePath -> IO ()
```
ksaric: Please add a comment where you copied this from and why.

src/Cardano/Shell/Update/Types.hs
```diff
@@ -3,8 +3,7 @@ module Cardano.Shell.Update.Types where
 import           Cardano.Prelude
 
 import qualified Data.Map as M
-
-import           Test.QuickCheck
+import           Test.QuickCheck (Gen, choose, frequency, listOf1)
```
ksaric: Thanks, better.

stack.yaml
```diff
@@ -36,6 +36,5 @@ extra-deps:
   - containers-0.5.11.0
   - libsystemd-journal-1.4.4
 
-
```
ksaric: :+1:

src/Cardano/Shell/Update/Lib.hs
HirotoShioi: Tried, but compiler complained so its not possible.

src/Cardano/Shell/Update/Lib.hs
HirotoShioi: I think this is optional since this field is not required when you're on Mac/Linux. You can have it `mempty` but that's kind of awkward.

src/Cardano/Shell/Update/Lib.hs
```diff
@@ -0,0 +1,136 @@
...
+    , udArchivePath :: Maybe FilePath
+    }
+
+-- Windows: https://github.com/input-output-hk/daedalus/blob/develop/installers/dhall/win64.dhall#L32-L35
+-- MacOS: https://github.com/input-output-hk/daedalus/blob/develop/installers/dhall/macos64.dhall#L31-L34
+-- Linux: https://github.com/input-output-hk/daedalus/blob/develop/installers/dhall/linux64.dhall#L29-L32
+updaterData :: UpdaterData
+updaterData = case buildOS of
+    Windows -> UpdaterData
+        "Installer.exe"
```
ksaric: Please indent these _after_ the constructor.
```
UpdaterData
    "Installer.exe"
    ...
```

app/Launcher/Main.hs
```diff
@@ -38,6 +39,8 @@ import           Data.X509.Extra (failIfReasons, genRSA256KeyPair,
 main :: IO ()
 main = do
 
+  -- _ <- runUpdater updaterData
```
ksaric: Remove this.
We need a simple `FilePath` when somebody wants to call the `runUpdater` - then it will try to run the installer on the `FilePath`.

src/Cardano/Shell/Update/Lib.hs
```diff
@@ -0,0 +1,145 @@
...
+#ifdef mingw32_HOST_OS
+import           System.Win32.Process (getCurrentProcessId)
+#endif
+
+-- | Updater path, args, windows runner path, archive path
+data UpdaterData = UpdaterData
+    { udPath        :: !FilePath
+    , udArgs        :: ![Text]
+    , udWindowsPath :: Maybe FilePath
+    , udArchivePath :: Maybe FilePath
```
ksaric: You don't need this since you know whether it's a Windows platform or not.
You just need one of these.

src/Cardano/Shell/Update/Lib.hs
```diff
@@ -0,0 +1,145 @@
...
+-- After it being generated, you run that script.
+runUpdater :: UpdaterData -> IO (Either UpdateError ExitCode)
+runUpdater = runUpdater' runCmd
+  where
+    runCmd :: FilePath -> [String] -> FilePath -> IO ExitCode
+    runCmd path args archive =
+        withCreateProcess (proc path (args <> [archive]))
+            $ \_in _out _err ph -> waitForProcess ph
+
+type RunCmdFunc = FilePath -> [String] -> FilePath -> IO ExitCode
```
ksaric: Please inline this in the function type.

src/Cardano/Shell/Update/Lib.hs
ksaric: You know whether you are on `Max/Linux` or `Windows` so you don't need another check for that.

Issue #229, opened 2019-07-18
Add Wallet and TLS specific configuration options using CLI


ksaric:
Continuation from https://github.com/input-output-hk/cardano-shell/issues/223 for the Wallet and TLS specific part of the configuration.

PR #232, opened 2019-07-22
[GH-229] Add Wallet and TLS specific configuration options using CLI.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/229


src/Cardano/Shell/Constants/CLI.hs
```diff
@@ -194,3 +198,76 @@ configBlockCLIParser =
...
+    PartialTLS
+        <$> lastOption configCertificateCLIParser
+        <*> lastOption configCertificateCLIParser
+        <*> lastOption configCertificateCLIParser
+
+-- | Certificate CLI parser.
+configCertificateCLIParser :: Parser PartialCertificate
+configCertificateCLIParser =
+    PartialCertificate
+        <$> option auto
```
deepfire: We're not allowing defaulting here (`lastOption`)?

ksaric: Yep, turns out I need to fix a couple of things...

ksaric: Turns out those "couple of things" are an issue.
Need to rewrite the config and we need to keep the configuration above two levels deep if we want to use this "mechanism".

ksaric:
@deepfire merged this since it's required for https://github.com/input-output-hk/cardano-shell/pull/236

PR #234, opened 2019-07-26
Core:  + coPBftSigThd; optionalise coStaticKeySigningKeyFile & coStaticKey

deepfire:
1. add `coPBftSigThd` to `Core`
2. make `coStaticKeySigningKeyFile` and `coStaticKeyDlgCertFile` optional

### DIFF
--- a/src/Cardano/Shell/Constants/CLI.hs
+++ b/src/Cardano/Shell/Constants/CLI.hs
@@ -116,18 +116,20 @@ configCoreCLIParser = PartialCore
           <> metavar "GENESIS-HASH"
           <> help "The genesis hash value."
            )
-    <*> lastStrOption
-           ( long "signing-key"
-          <> metavar "FILEPATH"
-          <> help "Path to the signing key."
-           )
-    <*> lastStrOption
-           ( long "delegation-certificate"
-          <> metavar "FILEPATH"
-          <> help "Path to the delegation certificate."
-           )
+    <*> ((Just <$>) <$> lastStrOption
+         ( long "signing-key"
+                <> metavar "FILEPATH"
+                <> help "Path to the signing key."
+                ))
+    <*> ((Just <$>) <$> lastStrOption
+         ( long "delegation-certificate"
+                <> metavar "FILEPATH"
+                <> help "Path to the delegation certificate."
+                ))
     <*> lastOption configNetworkMagicCLIParser
     <*> lastOption configDBVersionCLIParser
+    <*> lastOption (option auto $ long "pbft-signature-threshold" <> metavar "DOUBLE"
+                    <> help "The PBFT signature threshold.")
   where
     -- | Parser for the network magic options.
     configNetworkMagicCLIParser :: Parser RequireNetworkMagic
--- a/src/Cardano/Shell/Constants/CLI.hs
+++ b/src/Cardano/Shell/Constants/CLI.hs
@@ -116,20 +116,26 @@ configCoreCLIParser = PartialCore
           <> metavar "GENESIS-HASH"
           <> help "The genesis hash value."
            )
-    <*> ((Just <$>) <$> lastStrOption
-         ( long "signing-key"
-                <> metavar "FILEPATH"
-                <> help "Path to the signing key."
-                ))
-    <*> ((Just <$>) <$> lastStrOption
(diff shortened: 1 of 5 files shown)


ksaric:
Please rebase on top of https://github.com/input-output-hk/cardano-shell/pull/237 since that will be the cleaner solution.

deepfire:
@ksaric, I can't just yet, for demo purposes.

deepfire:
@ksaric, updated.

src/Cardano/Shell/Constants/CLI.hs
```diff
@@ -116,18 +116,20 @@ configCoreCLIParser = PartialCore
...
-           ( long "signing-key"
-          <> metavar "FILEPATH"
-          <> help "Path to the signing key."
-           )
-    <*> lastStrOption
-           ( long "delegation-certificate"
-          <> metavar "FILEPATH"
-          <> help "Path to the delegation certificate."
-           )
+    <*> ((Just <$>) <$> lastStrOption
```
ksaric: To be perfectly honest, I don't see the point of doing this.
Think about it, we already have a `Maybe`.

src/Cardano/Shell/Constants/CLI.hs
```diff
@@ -116,18 +116,20 @@ configCoreCLIParser = PartialCore
...
+                <> help "Path to the signing key."
+                ))
+    <*> ((Just <$>) <$> lastStrOption
+         ( long "delegation-certificate"
+                <> metavar "FILEPATH"
+                <> help "Path to the delegation certificate."
+                ))
     <*> lastOption configNetworkMagicCLIParser
     <*> lastOption configDBVersionCLIParser
+    <*> lastOption (option auto $ long "pbft-signature-threshold" <> metavar "DOUBLE"
```
ksaric: Should use existing functions.

src/Cardano/Shell/Constants/PartialTypes.hs
```diff
@@ -49,17 +49,12 @@ data PartialCardanoConfiguration = PartialCardanoConfiguration
...
-    -- ^ Static key signing file.
-    , pcoStaticKeyDlgCertFile       :: !(Last FilePath)
-    -- ^ Static key delegation certificate.
+    , pcoStaticKeySigningKeyFile    :: !(Last (Maybe FilePath))
+    , pcoStaticKeyDlgCertFile       :: !(Last (Maybe FilePath))
     , pcoRequiresNetworkMagic       :: !(Last RequireNetworkMagic)
-    -- ^ Do we require the network byte indicator for mainnet, testnet or staging?
     , pcoDBSerializeVersion         :: !(Last Integer)
-    -- ^ Versioning for values in node's DB.
+    , pcoPBftSigThd                 :: !(Last (Maybe Double))
```
ksaric: This just looks weird.

dcoutts: Yeah it doesn't need to be this way. It's already optional. It just means when we finalise the config, it's ok if it has no default value, we convert it into a `Maybe` at that point. Feel free to follow-up to fix this.

ksaric: Will do.

deepfire: The idea was to have a `Maybe` in the final `CardanoConfiguration`.  Now I'm struggling to remember why I thought it was needed.

deepfire: Undoing in https://github.com/input-output-hk/cardano-shell/pull/246.

src/Cardano/Shell/Constants/CLI.hs
deepfire: @ksaric, consider that we need to detect if the end user supplied just one of the certificate and the key -- that's an error.

If we don't have the `Maybe`, we wouldn't be able to detect it.

src/Cardano/Shell/Constants/PartialTypes.hs
deepfire: Yes, I remember now!

https://github.com/input-output-hk/cardano-shell/pull/246#issuecomment-518907922 is the explanation for `Maybe Double` in `coPbftSigThd`!

Issue #242, opened 2019-08-01
Add missing CLI options from `cardano-node` to `cardano-shell`


ksaric:
There are some missing CLI options that exist in `cardano-node` but not in `cardano-shell`:
```
data CLI = CLI {
    cliCoreNodeId   :: CoreNodeId,
    cliNumCoreNodes :: NumCoreNodes,
    cliProtocol     :: Protocol,
    cliCommon       :: CommonCLI
  }
```

We should add them in `cardano-shell`.

PR #243, opened 2019-08-01
[GH-242] Add missing CLI options to shell.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/242

Can be run using `stack build --fast -j12 --nix && stack exec cardano-shell-exe --nix -- --log-config configuration/log-configuration.yaml --system-start 123456789 --slot-length 123456789` for example.

Also fixing https://github.com/input-output-hk/cardano-shell/pull/234/files/51c691aff16e719c94f3ad85b938094d6cba9185#diff-6111c9baeacdb1421faa3d1f338c9728

### DIFF
--- a/src/Cardano/Shell/Configuration/Lib.hs
+++ b/src/Cardano/Shell/Configuration/Lib.hs
@@ -34,7 +34,6 @@ import           Cardano.Shell.Constants.PartialTypes (PartialBlock (..), Partia
                                                        PartialLastKnownBlockVersion (..),
                                                        PartialNTP (..),
                                                        PartialNode (..),
-                                                       PartialSSC (..),
                                                        PartialTLS (..),
                                                        PartialTXP (..),
                                                        PartialUpdate (..),
@@ -44,8 +43,8 @@ import           Cardano.Shell.Constants.Types (Block (..),
                                                 Certificate (..), Core (..),
                                                 DLG (..),
                                                 LastKnownBlockVersion (..),
-                                                NTP (..), Node (..), SSC (..),
-                                                TLS (..), TXP (..), Update (..),
+                                                NTP (..), Node (..), TLS (..),
+                                                TXP (..), Update (..),
                                                 Wallet (..))
 
 -- | Converting a @Last@ to an @Either@
@@ -73,14 +72,10 @@ finaliseCardanoConfiguration PartialCardanoConfiguration{..} = do
     ccNTP                    <- finaliseNTP pccNTP
     ccUpdate                 <- finaliseUpdate pccUpdate
     ccTXP                    <- finaliseTXP pccTXP
-    ccSSC                    <- finaliseSSC pccSSC
     ccDLG                    <- finaliseDLG pccDLG
     ccBlock                  <- finaliseBlock pccBlock
-
     ccNode                   <- finaliseNode pccNode
-
     ccTLS                    <- finaliseTLS pccTLS
-
     ccWallet                 <- finaliseWallet pccWallet
 
     pure CardanoConfiguration{..}
@@ -95,11 +90,8 @@ finaliseCardanoConfiguration PartialCardanoConfiguration{..} = do
         coGenesisHash                   <- lastToEither "Unspecified coGenesisHash"
                                             pcoGenesisHash
 
-        coNodeId                        <- lastToEither "Unspecified coNodeId"
-                                            pcoNodeId
-
-        coNumCoreNodes                  <- lastToEither "Unspecified coNumCoreNodes"
-                                            pcoNumCoreNodes
+        let coNodeId                    = getLast pcoNodeId
+        let coNumCoreNodes              = getLast pcoNumCoreNodes
 
         coNodeProtocol                  <- lastToEither "Unspecified coNodeProtocol"
                                             pcoNodeProtocol
@@ -109,9 +101,6 @@ finaliseCardanoConfiguration PartialCardanoConfiguration{..} = do
 
         coRequiresNetworkMagic          <- lastToEither "Unspecified coRequiresNetworkMagic"
                                             pcoRequiresNetworkMagic
-
-        coDBSerializeVersion            <- lastToEither "Unspecified coDBSerializeVersion"
-                                            pcoDBSerializeVersion
 
         let coPBftSigThd                = getLast pcoPBftSigThd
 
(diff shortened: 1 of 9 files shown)


src/Cardano/Shell/Configuration/Lib.hs
```diff
@@ -89,16 +89,31 @@ finaliseCardanoConfiguration PartialCardanoConfiguration{..} = do
...
+                                            pcoGenesisHash
 
-        coRequiresNetworkMagic      <- lastToEither "Unspecified coRequiresNetworkMagic"    pcoRequiresNetworkMagic
-        coDBSerializeVersion        <- lastToEither "Unspecified coDBSerializeVersion"      pcoDBSerializeVersion
+        coNodeId                        <- lastToEither "Unspecified coNodeId"
+                                            pcoNodeId
 
-        coPBftSigThd                <- lastToEither "Unspecified coPBftSigThd"              pcoPBftSigThd
+        coNumCoreNodes                  <- lastToEither "Unspecified coNumCoreNodes"
+                                            pcoNumCoreNodes
```
dcoutts: Ah, this can't be a required param. It turns out that this param is only used for some protocols, and not the production PBFT one. It's used only for the simple demo protocols.

I think the right thing to do is to make `coNumCoreNodes` itself by a `Maybe` and treat it like we do for `pcoStaticKeySigningKeyFile` using `getLast`.

src/Cardano/Shell/Configuration/Lib.hs
```diff
@@ -89,16 +89,31 @@ finaliseCardanoConfiguration PartialCardanoConfiguration{..} = do
...
+                                            pcoNodeProtocol
+
+        let coStaticKeySigningKeyFile   = getLast pcoStaticKeySigningKeyFile
+        let coStaticKeyDlgCertFile      = getLast pcoStaticKeyDlgCertFile
+
+        coRequiresNetworkMagic          <- lastToEither "Unspecified coRequiresNetworkMagic"
+                                            pcoRequiresNetworkMagic
+
+        coDBSerializeVersion            <- lastToEither "Unspecified coDBSerializeVersion"
+                                            pcoDBSerializeVersion
```
dcoutts: I don't think the new node makes us of such a param. You can double check, but I think we'll be able to drop this.

src/Cardano/Shell/Configuration/Lib.hs
```diff
@@ -107,17 +122,20 @@ finaliseCardanoConfiguration PartialCardanoConfiguration{..} = do
     finaliseTXP :: PartialTXP -> Either Text TXP
     finaliseTXP PartialTXP{..} = do
 
-        txpMemPoolLimitTx           <- lastToEither "Unspecified txpMemPoolLimitTx"         ptxpMemPoolLimitTx
-        txpAssetLockedSrcAddress    <- lastToEither "Unspecified txpAssetLockedSrcAddress"  ptxpAssetLockedSrcAddress
+        txpMemPoolLimitTx           <- lastToEither "Unspecified txpMemPoolLimitTx"
+                                        ptxpMemPoolLimitTx
+
+        txpAssetLockedSrcAddress    <- lastToEither "Unspecified txpAssetLockedSrcAddress"
+                                        ptxpAssetLockedSrcAddress
```
dcoutts: This will not be used either and can be dropped.

src/Cardano/Shell/Constants/Types.hs
```diff
@@ -259,21 +274,25 @@ data Block = Block
...
+    -- ^ Block retrieval queue capacity.
+    , noPendingTxResubmissionPeriod     :: !Int
+    -- ^ Minimal delay between pending transactions resubmission.
+    , noWalletProductionApi             :: !Bool
+    -- ^ Whether hazard wallet endpoint should be disabled.
+    , noWalletTxCreationDisabled        :: !Bool
+    -- ^ Disallow transaction creation or re-submission of
+    -- pending transactions by the wallet.
+    , noExplorerExtendedApi             :: !Bool
+    -- ^ Enable explorer extended API for fetching more.
```
dcoutts: I don't think these last 3 will ever be used in the new node.

src/Cardano/Shell/Constants/Types.hs
```diff
@@ -259,21 +274,25 @@ data Block = Block
...
+    , noSlotLength                      :: !Integer
+    -- ^ Slot length time.
+    , noNetworkConnectionTimeout        :: !Int
+    -- ^ Network connection timeout in milliseconds.
+    , noConversationEstablishTimeout    :: !Int
+    -- ^ Conversation acknowledgement timeout in milliseconds.
+    , noBlockRetrievalQueueSize         :: !Int
+    -- ^ Block retrieval queue capacity.
+    , noPendingTxResubmissionPeriod     :: !Int
+    -- ^ Minimal delay between pending transactions resubmission.
```
dcoutts: Check with the network team about all these 4 params. The set of tunables will be different from the old node.

dcoutts:
Oh and looking again, we also do not need `SSC`/`PartialSSC` at all. That applies only to Ouroboros classic.

src/Cardano/Shell/Constants/Types.hs
coot: * `noConversationEstablishTimeout` - I don't think we need it; we can relay on TCP connection timeouts. @dcoutts I don't this we need to timeout `connect`, this only complicates things, and it is configurable on the system level.
* `noConversationEstablishTimeout` - we'll need something like this for the handshake protocol, let's call it `noHandshakeTimeout`.  We'll use it for both sides of the handshake protocol: waiting for the initial message and waiting for the replay.
* `noBlockRetrievalQueueSize` - not needed
* `noPendingTxResubmissionPeriod` - not needed

src/Cardano/Shell/Constants/Types.hs
```diff
@@ -226,21 +222,9 @@ data LastKnownBlockVersion = LastKnownBlockVersion
...
-    , sscNoReportNoSecretsForEpoch1    :: !Bool
-      -- ^ Don't print “SSC couldn't compute seed” for the first epoch.
-    } deriving (Eq, Show)
-
 data TXP = TXP
     { txpMemPoolLimitTx        :: !Int
       -- ^ Limit on the number of transactions that can be stored in the mem pool.
-    , txpAssetLockedSrcAddress :: ![Text]
-      -- ^ Set of source address which are asset-locked. Transactions which
-      -- use these addresses as transaction inputs will be silently dropped.
```
dcoutts: Turns out this one is still needed for the Byron era. So this and all the other places above where this gets processed.

PR #246, opened 2019-08-06
pcoPBftSigThd:  undo the Maybe

deepfire:
1. ~Undoes the thinko~ There's some controversy -- apparently the change was motivated -- see the discussion below.
2. ~Fixes~ disables the perpetual `test-ipc`/Darwin ~issue~ test, that looked like so:
```
dns.js:246
    this._handle = new ChannelWrap();
                   ^

Error: EFILE
    at new Resolver (dns.js:246:20)
    at dns.js:377:23
    at NativeModule.compile (bootstrap_node.js:600:7)
    at NativeModule.require (bootstrap_node.js:545:18)
    at net.js:45:13
    at NativeModule.compile (bootstrap_node.js:600:7)
    at NativeModule.require (bootstrap_node.js:545:18)
    at createWritableStdioStream (internal/process/stdio.js:163:17)
    at process.getStdout [as stdout] (internal/process/stdio.js:14:14)
    at console.js:249:38
builder for '/nix/store/8hl7r13yw4sbmyylcdpq2q3qj8d12jmk-test-ipc-x86_64-darwin.drv' failed with exit code 1
```

### DIFF
--- a/src/Cardano/Shell/Presets.hs
+++ b/src/Cardano/Shell/Presets.hs
@@ -38,7 +38,7 @@ mainnetConfiguration =
           , pcoStaticKeyDlgCertFile     = pure Nothing
           , pcoRequiresNetworkMagic     = pure NoRequireNetworkMagic
           , pcoDBSerializeVersion       = pure 0
-          , pcoPBftSigThd               = pure Nothing
+          , pcoPBftSigThd               = pure 0.22
           }
     , pccNTP =
         PartialNTP
@@ -156,7 +156,7 @@ devConfiguration =
           , pcoStaticKeyDlgCertFile     = pure Nothing
           , pcoRequiresNetworkMagic     = pure RequireNetworkMagic
           , pcoDBSerializeVersion       = pure 0
-          , pcoPBftSigThd               = pure Nothing
+          , pcoPBftSigThd               = pure 0.22
           }
     , pccNTP =
         PartialNTP
--- a/test.nix
+++ b/test.nix
@@ -21,4 +21,8 @@ let
       touch $out
     '';
   };
-in builtins.listToAttrs (map mkTest [ "x86_64-linux" "x86_64-darwin" ])
+in builtins.listToAttrs (map mkTest [
+   "x86_64-linux"
+   # "x86_64-darwin" -- This has been (approximately since forever) causing intermittent failures, such as:
+   # https://iohk-nix-cache.s3-eu-central-1.amazonaws.com/log/8hl7r13yw4sbmyylcdpq2q3qj8d12jmk-test-ipc-x86_64-darwin.drv
+   ])
--- a/src/Cardano/Shell/Constants/CLI.hs
+++ b/src/Cardano/Shell/Constants/CLI.hs
@@ -130,7 +130,7 @@ configCoreCLIParser = PartialCore
         )
     <*> lastOption configNetworkMagicCLIParser
     <*> lastOption configDBVersionCLIParser
-    <*> (Just <<$>> lastDoubleOption
+    <*> (lastDoubleOption
            ( long "pbft-signature-threshold"
           <> metavar "DOUBLE"
           <> help "The PBFT signature threshold."
--- a/src/Cardano/Shell/Constants/PartialTypes.hs
+++ b/src/Cardano/Shell/Constants/PartialTypes.hs
@@ -54,7 +54,7 @@ data PartialCore = PartialCore
     , pcoStaticKeyDlgCertFile       :: !(Last (Maybe FilePath))
     , pcoRequiresNetworkMagic       :: !(Last RequireNetworkMagic)
     , pcoDBSerializeVersion         :: !(Last Integer)
-    , pcoPBftSigThd                 :: !(Last (Maybe Double))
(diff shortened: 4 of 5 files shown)


src/Cardano/Shell/Presets.hs
```diff
@@ -38,7 +38,7 @@ mainnetConfiguration =
           , pcoStaticKeyDlgCertFile     = pure Nothing
           , pcoRequiresNetworkMagic     = pure NoRequireNetworkMagic
           , pcoDBSerializeVersion       = pure 0
-          , pcoPBftSigThd               = pure Nothing
+          , pcoPBftSigThd               = pure 0.22
```
erikd: Would be nice to have a comment about what that value is and why its there, even just a pointer to elsewhere would be a huge improvement.

src/Cardano/Shell/Presets.hs
```diff
@@ -156,7 +156,7 @@ devConfiguration =
           , pcoStaticKeyDlgCertFile     = pure Nothing
           , pcoRequiresNetworkMagic     = pure RequireNetworkMagic
           , pcoDBSerializeVersion       = pure 0
-          , pcoPBftSigThd               = pure Nothing
+          , pcoPBftSigThd               = pure 0.22
```
erikd: Since the same value occurs in two places, why not make that a top level value and then have a comment on that value explaining what it is and why it is there.

src/Cardano/Shell/Presets.hs
erikd: Also, why are all those `pure`s there on every field?

erikd: Oh, those `pure` are required by the use of `Last`? Ok, suppose so.

src/Cardano/Shell/Presets.hs
erikd: More generally, there is a *lot* of duplication in that file. Would it not be better so defined one of them, and then define the other as a set of modifications applied to that?

src/Cardano/Shell/Presets.hs
deepfire: Yes, the `pure`'s are due to `Last`, indeed.

deepfire:
Yes!

I remember now why we need `Maybe` in `pcoPBftSigThd`!

It's because of https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/src/Ouroboros/Consensus/Node/ProtocolInfo/Byron.hs#L133

We _need_ the maybe, iff we're to enforce this last-step check in `protocolInfoByron`.

erikd:
Suggests we also need a comment about why we need the `Maybe`.

test.nix
```diff
@@ -21,4 +21,8 @@ let
       touch $out
     '';
   };
-in builtins.listToAttrs (map mkTest [ "x86_64-linux" "x86_64-darwin" ])
+in builtins.listToAttrs (map mkTest [
+   "x86_64-linux"
+   # "x86_64-darwin" -- This has been (approximately since forever) causing intermittent failures, such as:
```
disassembler: I approve removing this from CI

deepfire: Danke schoen, @disassembler!

dcoutts:
We do *not* need the maybe, because the `Last` can already be converted to the `Maybe`, turning the default no answer of `Last` into the `Maybe` that later gets interpreted as the default.

Or we can move the default to the top level set of defaults.

We only need a `Maybe` if we need to be able to override on the command line to reset something back to empty, which is not the case here. But to imagine that kind of example, think of something like `--use-blah --use-empty-blah` where the second one resets to the empty value. That would need a `Last (Maybe Blah)`.

ksaric:
Fixed in https://github.com/input-output-hk/cardano-shell/pull/243

dcoutts:
> Fixed in #243

Yes, this PR does it right. I was just explaining why.

ksaric:
@deepfire cherry-picked the node-ipc tests fix into https://github.com/input-output-hk/cardano-shell/pull/243

ksaric:
This can be closed.

ksaric:
Reverting back the commit - https://buildkite.com/input-output-hk/cardano-shell/builds/822#e50d1d35-7d7c-44f7-a041-2d6d7c1806f8/949-1250

deepfire:
> We do _not_ need the maybe, because the `Last` can already be converted to the `Maybe`, turning the default no answer of `Last` into the `Maybe` that later gets interpreted as the default.

But we would break the abstraction, by referring to the `PartiaCardanolConfiguration`'s value, and that would lead to presumably code coupling, no?

The over-arching intent of the approach seems to be to keep the final `CardanoConfiguration` as the source of truth.  And so if the consumer (the `ouroboros-network` code I linked) wants a `Maybe`, then, ostensibly, that `Maybe` is a part of configuration we want?

> Or we can move the default to the top level set of defaults.

Yes, that was the other option.. I frankly don't recall why we made the decision to put the defaulting in `ouroboros-network`.

> We only need a `Maybe` if we need to be able to override on the command line to reset something back to empty, which is not the case here. 

Oh, but isn't the opposite also true?  We may plausibly want to override the empty default to a non-empty?

Issue #238, opened 2019-07-29
Move the configuration to the `cardano-node`


ksaric:
The plan is that somebody moves the config types that are specific to the node out of the `cardano-shell` and into the `cardano-node`. The principle is that the `shell` is a general-purpose library, and the `node` is a specific user of the library.
The types were originally incubated in the `shell` because there wasn't anywhere else really to put them at the time.

The `cardano-ledger` shouldn't depend on `monitoring` which depends on `cardano-shell`.
The `cardano-ledger` only needs to depend on `contra-tracer` as the `network` and `consensus` do.

PR #250, opened 2019-08-08
[GH-238] Move the configuration to the cardano-node.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/238

:sob:


ksaric:
@CodiePP I would like to merge this since we have the git history and we do need this version in order to see the upstream projects that depend on the cardano-shell compile without module dependencies that are contained in the previous version of cardano-shell.

src/Cardano/Shell/DaedalusIPC.hs
```diff
@@ -56,24 +55,25 @@ instance ToJSON MsgOut where
...
-    -- ^ Logging object
-    -> Int
+    :: Int
     -- ^ Port number to send to Daedalus
     -> IO ()
-daedalusIPC tr port = setupNodeChannel >>= \case
+daedalusIPC port = setupNodeChannel >>= \case
     Right chan -> do
-        logInfo tr "Daedalus IPC server starting"
+        putTextLn "Daedalus IPC server starting"
```
dcoutts: Perhaps it will make sense to use the `contratrace` `Tracer` abstraction here in future.

ksaric: The idea that I got was to do something like:
```
data LoggingFunctions = LoggingFunctions
    { logInfo :: () -> IO ()
    , ...
    }
```

And hide the implementation inside? We then don't even need that dependency if we want to push everything out.

Issue #270, opened 2019-09-05
Add simple tests for exit codes handling and add safe mode


ksaric:
We currently need a more robust way of testing how the code behaves when different exit codes are returned from Daedalus.
Also, we are missing the "safe mode" option.

PR #272, opened 2019-09-06
[GH-270] Add simple tests for exit codes handling and add safe mode.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/270

### DIFF
--- a/cardano-launcher/src/Cardano/Shell/Launcher.hs
+++ b/cardano-launcher/src/Cardano/Shell/Launcher.hs
@@ -7,19 +7,18 @@ module Cardano.Shell.Launcher
     , WalletPath (..)
     , ExternalDependencies (..)
     -- * Functions
-    , runWallet
+    , runWalletProcess
     -- * Critical exports (testing)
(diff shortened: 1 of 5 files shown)


cardano-launcher/src/Cardano/Shell/Launcher.hs
```diff
@@ -46,9 +56,12 @@ newtype UpdateRunner = UpdateRunner { runUpdate :: IO ExitCode }
 
 newtype LauncherRunner = LauncherRunner { runLauncher :: IO ExitCode }
```
HirotoShioi: I've overlooked this in the previous PR but, I think the naming is a bit misleading. We're only re-running part of the launcher code (which is `runWallet`) and not the whole launcher process.

HirotoShioi: LGTM on everything else. Well done.

ksaric: Good catch. Fixed.

Issue #271, opened 2019-09-06
Implement CLI for the launcher


HirotoShioi:
We want to have CLI for the launcher so we can specify the filepath to the launcher-config.yaml when running it.

PR #275, opened 2019-09-10
Implement CLI for the launcher

HirotoShioi:
https://github.com/input-output-hk/cardano-shell/issues/271

TODO: Merge https://github.com/input-output-hk/cardano-shell/pull/255 and rebase from master

How to run: `:main -c /path/to/config.yaml`

### DIFF
--- a/cardano-launcher/src/Cardano/Shell/CLI.hs
+++ b/cardano-launcher/src/Cardano/Shell/CLI.hs
@@ -24,7 +24,6 @@ newtype LauncherOptionPath = LauncherOptionPath
 --
 -- this will enable the launcher to load launcher-config.yaml from the same
 -- directory as the cardano-launcher binary
--- This will be used if the filepath was not provided via CLI
 getDefaultConfigPath :: IO FilePath
 getDefaultConfigPath = do
     launcherDir <- takeDirectory <$> getExecutablePath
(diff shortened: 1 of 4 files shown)


cardano-launcher/src/Cardano/Shell/CLI.hs
```diff
@@ -0,0 +1,60 @@
...
+newtype LauncherOptionPath = LauncherOptionPath
+    { getLauncherOptionPath  :: FilePath
+    }
+
+-- | Default path to the launcher-config.yaml file
+--
+-- This will be used if the filepath was not provided via CLI
+getDefaultConfigPath :: IO FilePath
+getDefaultConfigPath = do
+    launcherDir <- takeDirectory <$> getExecutablePath
```
ksaric: Are you sure about this one? Did you test this somehow? I presume that the executable will be provided once we create an executable out of it, but I'm still not sure which one.
```
$> stack ghci

> import           System.Environment (getExecutablePath)
> getExecutablePath
"/home/ksaric/.stack/programs/x86_64-linux/ghc-8.6.5/lib/ghc-8.6.5/bin/ghc"
```

HirotoShioi: This is used for fetching configuration file on both Mac and Windows.

ksaric: I understand that, what I'm asking is what path will be returned here if we run it using Nix, for example, and would that work?

cleverca22: that is why linux uses the `--config` argument on the command line, rather then expecting the cfg to be in the same dir as the binary

https://github.com/input-output-hk/daedalus/blob/develop/installers/nix/linux.nix#L47-L48

Issue #276, opened 2019-09-10
Supply context for the substitute environment variables


HirotoShioi:
Some of the values in the `launcher-config.yaml` file has a placeholder wrapped with brackets (e.g. `${FOO}` ).
https://github.com/input-output-hk/daedalus/blob/develop/installers/dhall/win64.dhall#L3

This needs to be replaced with the env var defined with same name. For example, if there's placeholder `${FOO}` then the launcher needs to lookup env var `FOO` and replace `${FOO}` with that value.

This is needed for launcher to understand:
- Where to store TLS certificates before launching Daedalus
- Where the state directory is (This is needed for storing the state of launch mode)
- Where to find update files

PR #281, opened 2019-09-19
Supply context for the substitute environment variables

HirotoShioi:
This PR introduces function `substituteEnvVars` which will traverse through yaml value and replace identifier text to actual value.
(e.g. if there was string value `${USER}`, it will look up environment variable name USER, then replace `${USER}` with that value)

https://github.com/input-output-hk/cardano-shell/issues/276

### DIFF
--- a/cardano-launcher/src/Cardano/Shell/Environment.hs
+++ b/cardano-launcher/src/Cardano/Shell/Environment.hs
@@ -1,0 +1,68 @@
+{-| This is an example of how you can perform env var substitution
+}
+
+{-# LANGUAGE OverloadedStrings #-}
+
+module Cardano.Shell.Environment where
+
+import           Cardano.Prelude
+
+import           Cardano.Shell.Launcher (LauncherOptions)
+import           Cardano.Shell.Template (substituteA)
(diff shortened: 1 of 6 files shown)


cardano-launcher/src/Cardano/Shell/Environment.hs
```diff
@@ -0,0 +1,69 @@
...
+  eSubstituted <- runExceptT $ do
+    decodedVal <- withExceptT (\e -> FailedToDecodeFile (show e)) $ 
+      ExceptT $ decodeFileEither yamlPath
+    substituted <- substituteEnvVars decodedVal
+    return substituted
+  case eSubstituted of
+    Left err -> return . Left $ err
+    Right sub -> case fromJSON sub of
+        Error _ -> return . Left $ FailedToParseLauncherOption
+        Success decoded -> return . Right $ decoded
```
HirotoShioi: This is just an example code I wrote to test that it actually works.
After this and https://github.com/input-output-hk/cardano-shell/pull/275 has been merged. I'll work on substituting the values the launcher actually needs

PR #284, opened 2019-09-20
Add cardano-launcher-test to hydra ci

HirotoShioi:
This PR would enable the hydra CI to run test suites for the `cardano-launcher`
https://hydra.iohk.io/build/1153187

Notice job `nix-tools.tests.cardano-launcher.cardano-launcher-test.x86_64-linux` has been added as one of the jobs.
Before: https://hydra.iohk.io/build/1153098#tabs-constituents
After: https://hydra.iohk.io/build/1153190#tabs-constituents

### DIFF
--- a/cardano-launcher/src/Cardano/Shell/Launcher.hs
+++ b/cardano-launcher/src/Cardano/Shell/Launcher.hs
@@ -158,16 +158,16 @@ handleDaedalusExitCode
     -> IO DaedalusExitCode
 handleDaedalusExitCode runUpdaterFunction restartWalletFunction = isoTo <<$>> \case
     RunUpdate               -> runUpdate runUpdaterFunction >> runRestart restartWalletFunction
-    -- ^ Run the actual update, THEN restart launcher.
+    -- Run the actual update, THEN restart launcher.
     -- Do we maybe need to handle the update ExitCode as well?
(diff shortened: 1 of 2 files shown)


cardano-launcher/src/Cardano/Shell/Launcher.hs
```diff
@@ -158,16 +158,16 @@ handleDaedalusExitCode
     -> IO DaedalusExitCode
 handleDaedalusExitCode runUpdaterFunction restartWalletFunction = isoTo <<$>> \case
     RunUpdate               -> runUpdate runUpdaterFunction >> runRestart restartWalletFunction
-    -- ^ Run the actual update, THEN restart launcher.
+    -- Run the actual update, THEN restart launcher.
```
erikd: Any reason for those? Turning Haddock docs into regular comments?

HirotoShioi: The reason why I've deleted those comments is because the hydra build was failing.
https://hydra.iohk.io/build/1153157/nixlog/1
I think this is because it's using haddock syntax within a function is not allowed

```haskell
-- ^ Test
testFoo :: Int -> Int -> Int
testFoo one two = (multiply one) + two
  where
     -- ^ This will make haddock build fail
     multiply = * 3
```

Issue #300, opened 2019-10-03
Update manual testing package script


HirotoShioi:
Due to the recent changes to the launcher, [nix-build script](https://github.com/input-output-hk/cardano-shell/blob/master/updater-test-win.nix) that was used to generate package for manual testing got out-dated. We should update it so that it works again.

PR #301, opened 2019-10-03
Update manual test script

HirotoShioi:
https://github.com/input-output-hk/cardano-shell/issues/300


HirotoShioi:
Yeah sorry for not having enough context. Basically the script was broken due to the recent changes (CLI, setting working directory, etc.) So this PR applies update to the script so that it'll work again.

Issue #308, opened 2019-10-10
Improve exception handling on TLS certificate generation


HirotoShioi:
Currently, TLS certificate generation function `generateTlsCertificates` will always exception, `CannotGenerateTLS` no matter how the function has failed. This makes debugging extremely hard since exception message does not tell you what the problem is.

Try to add more exception message such as configuration file not found, invalid key, etc.

HirotoShioi:
I'm trying to import this module, but the compiler is telling me it's behind hidden module.
https://github.com/input-output-hk/cardano-sl-x509/blob/master/src/Data/X509/Extra.hs#L31

PR #309, opened 2019-10-10
Exception handling upon certificate generation

HirotoShioi:
https://github.com/input-output-hk/cardano-shell/issues/308


cardano-launcher/test/LauncherSpec.hs
```diff
@@ -202,3 +207,58 @@ withSetEnvs path homePath action = bracket
         , "DAEDALUS_INSTALL_DIRECTORY"
         , "LAUNCHER_CONFIG"
         ]
+
+generateTLSCertSpec :: Spec
```
ksaric: Niiice, Hiroto.
Very nice.

:tada: 

You are using the `withSystemTempDirectory` function to test it, which is good in this case.
In the future, it would be slightly better (in some cases, this is good as it is) to actually provide a stub function where you generate the TLS - this would make the code more flexible.

cardano-launcher/app/Main.hs
```diff
@@ -136,7 +123,7 @@ data LauncherExceptions
     = CannotGenerateTLS
     | LauncherOptionsError Text
     | WorkingDirectoryDoesNotExist FilePath
-    deriving (Eq)
+    | FailedToGenerateTLS TLSError
```
ksaric: The `Eq` can be added here, I think. It can usually be useful to have `Eq` on types if possible.

Issue #310, opened 2019-10-11
Daedalus mock exits with ExitSuccess


HirotoShioi:
Mock the we're using to test the update system exits itself with `ExitSuccess` even though we want it to exit with `ExitFailure 20`.

PR #311, opened 2019-10-11
Make it so that daedalus mock fails with ExitFailure 20

HirotoShioi:
https://github.com/input-output-hk/cardano-shell/issues/310

Use `exitWith` from `System.Exit` so that mock exits itself with `ExitFailure 20` which is what we want.


HirotoShioi:
I've tested and it works as expected.

Issue #306, opened 2019-10-09
Simplify the handling of updates


ksaric:
We currently have two separate mechanisms for the update - one for Windows and one for Unix.
We can simplify the situation if we write a batch file for Windows every time we run the launcher and then we don't have to have a separate Windows runner.

PR #314, opened 2019-10-14
[GH-306] Added missing Data.Text import on Windows.

ksaric:
Fixing the missing import I missed.
I'm not sure how this passed CI.


ksaric:
Abusing my admin privilege since @HirotoShioi is off, a national holiday.

PR #321, opened 2019-10-16
automatically disable tls if no path is configured

cleverca22:
solves https://github.com/input-output-hk/cardano-shell/issues/328


HirotoShioi:
Please create an issue for this so we can understand what this solve.

Issue #340, opened 2019-10-31
Use osascript to dispaly error message on darwin


HirotoShioi:
Currently, cardano-launcher fails silently when executed on darwin which make it very unfriendly for users.
To deal with this, we can use `osascript` which can be used to display arbitrary message as popup message.

https://scriptingosx.com/2018/08/user-interaction-from-bash-scripts/

PR #341, opened 2019-10-31
Use 'osascript' to display message on Mac

HirotoShioi:
https://github.com/input-output-hk/cardano-shell/issues/340

### DIFF
--- a/cardano-launcher/app/Main.hs
+++ b/cardano-launcher/app/Main.hs
@@ -1,5 +1,7 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE CPP #-}
+
 
 module Main where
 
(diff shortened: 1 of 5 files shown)


cardano-launcher/app/Main.hs
```diff
@@ -277,3 +278,29 @@ silence :: IO a -> IO a
...
+-- | it
+-- | https://scriptingosx.com/2018/08/user-interaction-from-bash-scripts/
+logErrorMessage :: MonadIO m => Trace m Text -> Text -> m ()
+logErrorMessage tracer msg = do
+    Trace.logError tracer msg
+    case buildOS of
+        OSX -> liftIO $ displayErrorDarwin msg
+        _   -> return ()
+
+displayErrorDarwin :: Text -> IO ()
```
CodiePP: conditional compilation that this function is only available on Darwin?

Issue #134, opened 2019-05-06
T9.4.1 find example, adapt to our needs


CodiePP:
find example, adapt to our needs

PR #347, opened 2019-11-08
[GH-134] T9.4.1 find example, adapt to our needs.

ksaric:
https://github.com/input-output-hk/cardano-shell/issues/134

The basic version of the launcher behavior, including updates.


ksaric:
> you might want to indicate how to setup the "model" for executing the spec: Invariant = TypeOK?, Properties = Properties?

Yes, exactly. Use "Temporal formula" to be `Spec` and add `Properties` to "Properties".

image

> also, how is it verified that it passes through all states? is there a negative test for this?

This is something I spent quite some time on. It turns out, it's not that simple.
However, the current spec contains two interesting properties that are checked:
1. If the current state is the wallet normal mode, we will eventually end up in one of these states - "success", "failure", "update_mode", "wallet_safe_mode".
2. The current state will, eventually, go through the "wallet_normal_mode".

The same cannot be stated for "wallet_safe_mode" since there is no guarantee that it will at some point end up there. The best we can do is say that it will end up there eventually like the first property states.

Issue #339, opened 2019-10-30
Add the lock file and the support for the lock file.


ksaric:
We can currently run the launcher several times and have several instances running.
We need to add the lock file and the support for running a single instance **_ONLY_**.

ksaric:
We currently have the `lockfile` - https://github.com/input-output-hk/cardano-shell/blob/master/lockfile
We just need the support from the installer to include it in the installation.

@cleverca22 ?

ksaric:
We will just create the `lockfile` and get a lock on it, instead of packing it with the installation.

cleverca22:
the lock file should be in the state directory

with the current implementation, its impossible to open 2 clusters at once

PR #348, opened 2019-11-12
Store lockfile at statePath

HirotoShioi:
https://github.com/input-output-hk/cardano-shell/issues/339#issuecomment-550490262

This PR makes it so that lockfile is stored in `statePath`

### DIFF
--- a/cardano-launcher/configuration/launcher/launcher-config.demo.yaml
+++ b/cardano-launcher/configuration/launcher/launcher-config.demo.yaml
@@ -22,7 +22,7 @@ nodeArgs:
 - --wallet-address
 - 127.0.0.1:0
 - --update-with-package
-statePath: ${XDG_DATA_HOME}/Daedalus/mainnet
+statePath: .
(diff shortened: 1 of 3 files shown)


cardano-launcher/configuration/launcher/launcher-config.demo.yaml
```diff
@@ -22,7 +22,7 @@ nodeArgs:
 - --wallet-address
 - 127.0.0.1:0
 - --update-with-package
-statePath: ${XDG_DATA_HOME}/Daedalus/mainnet
+statePath: .
```
HirotoShioi: This config file is used to test the launcher on our local machine, modifying the value should not matter.

cleverca22: it may help to just make a custom launcher config for each testcase, rather then trying to use copies of the generated config

ksaric:
This needs to be considered when (QA) testing.

PR #352, opened 2019-11-13
update dependencies

CodiePP:
upgrade `iohk-monitoring`:
  should already clean up some problems
  like the message about not being able to remove the symlink.
  (see issue https://github.com/input-output-hk/iohk-monitoring-framework/issues/452)

upgrade `cardano-prelude`:
  to same revision that cardano-node uses


ksaric:
Works correctly, there are no issues with the logfile missing anymore.

PR #359, opened 2020-03-28
tlsConfig passed via launcherConfig

disassembler:
This PR changes tlsConfig to be an attribute in the launcher config instead of being a separate configuration file. We've tested it with daedalus flight and can confirm tls certificates are generated. This probably breaks tests.


cardano-launcher/app/Main.hs
```diff
@@ -77,7 +72,7 @@ main = silence $ do
 
     -- This function either stubs out the wallet exit code or
     -- returns the "real" function.
-    let walletExecutionFunction =
+    let walletExectionFunction =
```
CodiePP: why do you want to rename this function?

disassembler: it's a revert of the entire commit that's failing to work on windows. Fully agree that the other name is better, but I don't have the time or expertise to fix the failures we were getting on windows with this commit. I did a bisect and found this was the culprit.

PR #360, opened 2020-04-09
remove 2 safemode flags

cleverca22:


disassembler:
yes, flags have changed in newer electron version.

PR #361, opened 2020-04-20
allow persisting safemode flag

cleverca22:


cardano-launcher/src/Cardano/Shell/Launcher.hs
```diff
@@ -27,8 +29,11 @@ module Cardano.Shell.Launcher
 import           Cardano.Prelude hiding (onException)
 
 import           Prelude (Show (..))
+import           Data.Aeson (FromJSON, ToJSON(toJSON), genericParseJSON, genericToJSON, defaultOptions)
+import           Data.Yaml as Y
 import qualified System.Process as Process
 import           Turtle (system)
+import           Test.QuickCheck (Arbitrary (arbitrary), elements)
```
ksaric, line 36: Not sure why this is here, are you testing this?
Also, it would be simpler to write `Arbitrary (..)`.

cardano-launcher/src/Cardano/Shell/Launcher.hs
```diff
@@ -74,6 +79,29 @@ data WalletMode
     | WalletModeSafe
     deriving (Eq, Show)
 
+instance Arbitrary WalletMode where
+  arbitrary = elements [ WalletModeSafe, WalletModeNormal ]
```
ksaric, line 83: Yeah, this is weird, not sure why we have it here?

cardano-launcher/src/Cardano/Shell/Launcher.hs
```diff
@@ -74,6 +79,29 @@ data WalletMode
...
+
+instance FromJSON WalletMode where
+  parseJSON (Bool True) = pure WalletModeSafe
+  parseJSON (String "safe") = pure WalletModeSafe
+  parseJSON (String "normal") = pure WalletModeNormal
+  parseJSON _ = pure WalletModeNormal
+
```
ksaric, line 90: It should have a single representation, not map 5 types into this one?
If you want to have a `String`, use a `String`, otherwise use a `Bool`.
Don't mix and match, no need for that.

cardano-launcher/src/Cardano/Shell/Launcher.hs
```diff
@@ -125,6 +153,22 @@ instance Isomorphism DaedalusExitCode ExitCode where
...
+    Right value -> pure $ smcSafeMode value
+    Left _      -> pure WalletModeNormal
+
+saveSafeMode :: FilePath -> WalletMode -> IO ()
+saveSafeMode stateDir mode = do
+  let safeModeConfigFile = getSafeModeConfigPath stateDir
+  Y.encodeFile safeModeConfigFile $ SafeModeConfig mode
+
+getSafeModeConfigPath :: FilePath -> FilePath
+getSafeModeConfigPath stateDir = stateDir </> "safemode.yaml"
```
ksaric, line 170: I have to admit that this looks pretty weird. The state is being persisted into a `*.yaml` config?

ksaric:
@cleverca22 This is something I can get on board with if somebody has a really good reason for or some of the more experienced developers agree with you.

disassembler:
The issue is electron doesn't support underlying GPU's that don't have GPU accceleration by default. A parameter needs passed to electron when it starts to tell it to use safe mode. We absolutely want to persist this as we don't want the user to have to get the blank screen every time they open Daedalus.

disassembler:
Other way around. The user gets the blank screen not in safe mode if they don't have GPU acceleration. exit code 22 instructs the launcher to restart daedalus process with safe mode flag which then shows the UI correctly. However, prior to this PR, when the user exits daedalus and restarts, it does not start in safe mode and they get the blank screen again.

Note that FC5 release today is going to point at this branch. @cleverca22 can address the comments and make sure any interface changes end up in daedalus. We need to target getting this done before the mainnet daedalus release.

cardano-launcher/src/Cardano/Shell/Launcher.hs
cleverca22: the `Bool` stuff is backwards compatibility with the old launcher and `safemode.yaml`

cardano-launcher/src/Cardano/Shell/Launcher.hs
cleverca22: i used yaml instead of a flat file, so we have the option to add more fields in the future

disassembler:
bors ping

ksaric:
There is no bors here, just merge AFAIK.

cleverca22:
bors was just added

PR #365, opened 2020-07-02
Update dependencies and Hackage index-state

erikd:
The dependency update was needed in db-sync.
Biggest changes were due to updates to protolude (via cardano-prelude).


oneEdoubleD:
LGTM. Had a local build error with `stack build --fast --nix` but `stack build --fast` worked and it appears to be a local path error.

